PR #12, opened 2019-06-14
add support for Hexen-format maps (without ACS)

fabiangreffrath:


coelckers:
Since this doesn't convert the specials to anything usable I do not see much of a point to add this right now. It's a good start for a working implementation, though.

fabiangreffrath:
Well, at least it loads these maps now and doesn't crash anymore. It doesn't allow for much more than map exploration as it currently is, though, that's granted.

fabiangreffrath:
> Since this doesn't convert the specials to anything usable

Do you know of a Hexen-linedef-to-generalized-linedef converter available?

coelckers:
No, such a thing does not exist. The Hexen types are far more versatile than Boom's generalized types and to a large degree not compatible,

fabiangreffrath:
Hm, okay, but maybe we could translate the most basic types so that it is at least possible to e.g. open a door or use a teleporter?

kraflab:
@fabiangreffrath I think we can close this one. What do you think?

fabiangreffrath:
Oh, yes, I think this can be closed by now. 😅

PR #17, opened 2019-06-21
Remove nonstandard gamesdir variable

vilhelmgray:
The Autotools canonical method of configuring the program installation
directory is via the bindir variable (e.g. "--bindir=/usr/games/doom").

### DIFF
--- a/prboom2/src/Makefile.am
+++ b/prboom2/src/Makefile.am
@@ -8,12 +8,10 @@
 
 SUBDIRS = SDL POSIX MAC PCSOUND TEXTSCREEN MUSIC   
 
-gamesdir=$(prefix)/games
-
 if BUILD_SERVER
-games_PROGRAMS = prboom-plus prboom-plus-game-server
+bin_PROGRAMS = prboom-plus prboom-plus-game-server
 else
-games_PROGRAMS = prboom-plus
+bin_PROGRAMS = prboom-plus
 endif
 
 CFLAGS = @CFLAGS@ @SDL_CFLAGS@ @LIBZ_CFLAGS@


fabiangreffrath:
Yes, but this is a game. And games may go into `/usr/games` instead of `/usr/bin` by FHS specs:
http://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch04s03.html

I don't see any problem in keeping this.

coelckers:
Looks like a change for the sake of making a change, nothing more.

vilhelmgray:
> Yes, but this is a game. And games may go into `/usr/games` instead of `/usr/bin` by FHS specs:
> http://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch04s03.html
> 
> I don't see any problem in keeping this.

Unfortunately, `/usr/games` is an optional directory so systems are not required to support. This is a problem for some systems such as Gentoo Linux which has deprecated `/usr/games` and no longer allow installation to there: https://wiki.gentoo.org/wiki/Project:Quality_Assurance/Policies#Games

The primary issue with the current code is that the installation directory is not configurable -- `make install` will install the PrBoom+ executable into `$(prefix)/games` and there's no way to change that location; this effectively means that `make install` does not work on systems such as Gentoo Linux that have restrictions on `/usr/games` and such.

If switching to `${bindir}` is not appealing, what I can do alternatively is create a configuration option to select the installation directory for `gamesdir`. We can default it to `$(prefix)/games` so there's no behavior change from what users already expect, but still give users the power to configure the installation directory if they so need it.

Would this be a good approach to resolving this issue? If not, is there an alternative solution that you would prefer?

fabiangreffrath:
Or, you change it to use bindir, so it is configurable, but remain with usr/games as the default.

fabiangreffrath:
Apparently, it isn't really easy (or even possible?) to change the default for `bindir` in Autoconf away from ` '${exec_prefix}/bin'` to anything else. Since `gamedir` isn't configurable, but `bindir` is, the `/usr/games` people, i.e. Debian, could use the `./configure` call to set their installation destination, whereas all the others could leave it at the default.

This approach is still cleaner than your subsequent PR in https://github.com/coelckers/prboom-plus/pull/18/files and I can confirm that Chocolate Doom already does it very similarly.

vilhelmgray:
> This approach is still cleaner than your subsequent PR in https://github.com/coelckers/prboom-plus/pull/18/files and I can confirm that Chocolate Doom already does it very similarly.

Yes, I believe we should follow the Chocolate Doom solution and just rely on `bindir` by default. That allows people who want `/usr/games` (e.g. Debian) to easily configure it via `./configure --bindir="/usr/games"`, while allowing everyone else to just `make install` without worry.

For that reason, instead of merging https://github.com/coelckers/prboom-plus/pull/18, we should merge this PR here.

fabiangreffrath:
@coelckers Upon further investigation, I have changed my opinion. This is not only safe to apply, it's even useful.

PR #18, opened 2019-06-23
Use bindir to configure gamesdir

vilhelmgray:
Default bindir to `$(prefix)/games` and allow configuration of gamesdir
via bindir.

This should allow users to install PrBoom+ on systems that restrict access to the default `/usr/games` directory. See https://github.com/chocolate-doom/chocolate-doom/pull/635 for more background.

### DIFF
--- a/prboom2/configure.ac
+++ b/prboom2/configure.ac
@@ -52,6 +52,10 @@ then
 fi
 AM_WITH_DMALLOC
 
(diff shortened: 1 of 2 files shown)


vilhelmgray:
Resolved in https://github.com/coelckers/prboom-plus/pull/17

PR #20, opened 2019-06-26
add support for SMMU swirling flats

fabiangreffrath:


coelckers:
This will require some support code in the GL renderer - either a shader to do the warping or invalidation of the texture for each frame.

fabiangreffrath:
Ah, GL, I knew I missed something... :wink:

fabiangreffrath:
Wait a minute, how does it work for animated flats? Are they invalidated each frame?

coelckers:
They are simply different textures for each frame. The swirling needs to change the image of the same texture.

fabiangreffrath:
> invalidation of the texture for each frame.

This is pretty much what happens in the software renderer anyway, right?

fabiangreffrath:
@JadingTsunami Any idea how swirling flats could get implemented in the OpenGL renderer?

JadingTsunami:
> @JadingTsunami Any idea how swirling flats could get implemented in the OpenGL renderer?

Hmm, yes, I think a similar effect can be done. It would require tesselating the flat and applying the swirl algorithm to the vertices. It won't look exactly the same, but it would be close.

Performance on older machines could become an issue though. And there would have to be some draw distance limits to prevent huge fields of water from slowing down. The other option I can think of would be DuDv maps, but that won't look quite the same and requires a newer version of OpenGL.

The last and cheapest option would be a simple sinusoidal texture translation, which would not recreate the swirling but would look more like a slow sloshing back and forth. It has to be pretty slow or else it will look like a strange conveyor belt.

So in sum I think we could replicate something, at varying levels of performance cost vs. visual niceness.

fabiangreffrath:
This sounds all complicated to me. Is there anything that could be done with shaders, maybe? Or can't we just recreate the texture from the modified flat during each tic?

JadingTsunami:
> This sounds all complicated to me.

You are correct. :)

> Is there anything that could be done with shaders, maybe?

Hmm, you mean a custom vertex shader to do the tesselated vertex warping? Yes it is possible. I am not sure the performance boost would be huge (I think it depends on the scene) but I suppose it hasn't been tried before. The net rendered effect ends up the same of course. But yes a shader would be better overall for performance.

> Or can't we just recreate the texture from the modified flat during each tic?

Yes, this is possible. And perhaps the most consistent idea. Basically a preprocessed, precompiled, animated flat. In fact I think both renderers could do it the same way in that case. Precompile a set of warped flats from a base flat and then render each frame as with any animated flat.

fabiangreffrath:
> Precompile a set of warped flats from a base flat and then render each frame as with any animated flat.

In Crispy Doom I have unrolled the whole 1024 steps sequence into a single giant lookup table (size 16 MB), which should significantly speed up generation of the distorted texture.

https://github.com/fabiangreffrath/crispy-doom/blob/master/src/doom/r_swirl.c

Gustavo6046:
Well, @JadingTsunami why not have a list of "special textures" (which includes flats, duh!), where one single surface is broken down into a grid, and the vertices' UV coordinates are transformed (instead of the vertices themselves)?

The plan is to be able to use the texture-mapped result as a texture itself. Can this be done with OpenGL? Then, every frame you'd update and render every **used** special texture once, then when rendering swirling walls or flats, you'd just take the associated rendered texture handle to tile across the actual polygon being rendered between glBegin/glEnd.

JadingTsunami:
> Well, @JadingTsunami why not have a list of "special textures" (which includes flats, duh!), where one single surface is broken down into a grid, and the vertices' UV coordinates are transformed (instead of the vertices themselves)?

Hmm, this sounds like the precomputed idea to me. Basically, for swirling textures/flats, generate each frame as a preprocessing step, then use 1 texture per tic as with any animated texture/flat.

Did you mean something different?

Gustavo6046:
Er, that works good actually. Maybe even on load? :o

kraflab:
@fabiangreffrath shall we close this one? It's probably out of scope at this point.

fabiangreffrath:
Yes, this won't make it into PrBoom+ anytime soon.

PR #23, opened 2019-09-16
Fix macOS build with OpenGL

Rathori:
Fix configure.ac to properly detect and use OpenGL on macOS.


Rathori:
FYI, this was also merged into the main prboom+ repo on SourceForge, so maybe you'll wanna merge the changes from there instead of merging this PR.

PR #24, opened 2019-09-23
print a blinking "arrow" next to the currently highlighted menu item

fabiangreffrath:
This helps with orientation in the menus for WADs with a non-red HUD
font where Boom's own usual red-to-anything color translation tables
fail.


fabiangreffrath:
Alternative for very long lines, though less visible:
```
M_DrawString(x, y, color, "_");
```

PR #34, opened 2019-12-28
Fix compilation with msys2

SmileTheory:
So I noticed this didn't compile cleanly with msys2, and even after making the minimum changes not everything was enabled.

So this enables everything.

I'm not too familiar with autotools, so I'm open to suggestions as to how to make the `makefile.am` changes less horrible.

As well, I didn't test on linux or visual studio, so those might both be broken. :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,17 @@
+Makefile
+Makefile.in
+aclocal.m4
+autom4te.cache
+autotools
+config.h
+config.h.in
+config.log
+config.status
+configure
+.deps
+stamp-h1
+*.exe
+prboom.spec
 prboom2/ICONS/icons.aps
 prboom2/ICONS/prboom-plus.wad
 prboom2/VisualC6/Debug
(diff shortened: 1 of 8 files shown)


coelckers:
I'd prefer if this was reviews by someone more familiar with that toolchain.

coelckers:
I concur with that sentiment. Although it might put off some of the die hard Linuxers, all those legacy projects should go away, better have one that can handle all platforms at once.

fabiangreffrath:
> Although it might put off some of the die hard Linuxers,

It's not that CMake is non-Linux nor that Automake is Linux-only. But, if you only care about Linux, then maintaining a CMake build system is just at least as cumbersome as maintaining an Automake one. Anyway, I am fine with the switch. :wink:

coelckers:
If you only care about Linux, cross platform compatibility does not matter. Once it does, though, Automake becomes a liability. I've experienced this with ZDoom during various phases that the parallel projects were out of sync, I also experienced this at work where I have to deal with code that needs to be compiled on iOS, Windows and Linux (and where CMake is not an option) and I experienced it here.

If there is a single build system that has been properly set up, things get a lot easier, and then adding new files is a simple task, no matter what you use. Rewriting the configuration is not a frequent occurence and thus shouldn't be used as a measurement for convenience.

CMake surely isn't the ultimate savior here - for that its scripting language is too bad - but currently it's clearly the best overall option, regardless of what some people in the Linux community think.

And once I have it all set up and working I'm going to delete the old projects. Hopefully then it gets easier to work on this and maybe get some more contributors.

PR #37, opened 2019-12-30
Fix compilation with msys2 (cmake edition)

SmileTheory:
So let's try this again.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,10 +38,3 @@ prboom2/release
 /prboom2/VisualC2015/ReleaseServer/prboom_server.tlog
 /prboom2/VisualC2015/ReleaseServer
 /build*
-CMakeFiles
-CMakeCache.txt
-Makefile
-*.cmake
-config.h
-prboom-plus.wad
-*.exe
--- a/.gitignore
+++ b/.gitignore
@@ -38,3 +38,10 @@ prboom2/release
 /prboom2/VisualC2015/ReleaseServer/prboom_server.tlog
 /prboom2/VisualC2015/ReleaseServer
 /build*
+CMakeFiles
+CMakeCache.txt
+Makefile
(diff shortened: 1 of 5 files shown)


.gitignore
```diff
@@ -38,3 +38,10 @@ prboom2/release
 /prboom2/VisualC2015/ReleaseServer/prboom_server.tlog
 /prboom2/VisualC2015/ReleaseServer
 /build*
+CMakeFiles
+CMakeCache.txt
+Makefile
+*.cmake
+config.h
+prboom-plus.wad
+*.exe
```
alexey-lysiuk: Please use out-of-tree builds. The `build*` directories filter was added especially for this.

prboom2/src/CMakeLists.txt
```diff
@@ -312,6 +312,8 @@ target_include_directories(prboom-plus PRIVATE
 target_link_libraries(prboom-plus PRIVATE
     ${OPENGL_LIBRARIES}
     ${SDL2_LIBRARIES}
+    winmm
+    comctl32
```
alexey-lysiuk: This is not the right place for Windows specific libraries. As you can see, it broke Linux and macOS targets.

SmileTheory: https://github.com/coelckers/prboom-plus/pull/37/commits/780c682cc2de4056a71816f4958cfe78dec47ff5#diff-ec649d8d689ac310801f57cae7a3f446R316

.gitignore
SmileTheory: https://github.com/coelckers/prboom-plus/pull/37/commits/780c682cc2de4056a71816f4958cfe78dec47ff5#diff-a084b794bc0759e7a6b77810e01874f2L41

Issue #14, opened 2019-06-14
please update the OPL emulator to the latest revision from Chocolate Doom


fabiangreffrath:
The OPL emulatior implementation in Chocolate Doom has been replaced with Nuked OPL3 v1.8 for better performance and accuracy. It might be a good idea to do the same here.

Clownacy:
Better performance? Aren't the Nuked emulators meant to be cycle-accurate, made by reimplementing synthesiser die-shots in C? Nuked OPN2 was added to Genesis Plus GX a while back, where it was by-far the slowest audio emulation backend.

PR #38, opened 2019-12-30
Replace DOSBox OPL2 emulator with Nuked OPL3 v1.8.

SmileTheory:
https://github.com/coelckers/prboom-plus/issues/14


SmileTheory:
I've been messing around with this, and I should mention that on my machine (core i5-2500k), nuked opl3 is not performant enough to run at a samplerate of 48000hz without glitching unless the build is made with `CMAKE_BUILD_TYPE="release"`, typically taking ~15ms per 43ms of audio.

coelckers:
Closed for the same reason as the other PR - with my current plans it makes not much sense to change the sound system as the code is to be removed anyway.

PR #43, opened 2020-01-10
Define _GNU_SOURCE to find sched_setaffinity

38-vita-38:
[man page](https://linux.die.net/man/2/sched_setaffinity):
> #define _GNU_SOURCE             /* See feature_test_macros(7) */
> #include <sched.h>
> int sched_setaffinity(pid_t pid, size_t cpusetsize, cpu_set_t *mask);

With HAVE_SCHED_SETAFFINITY prb+ release build runs a bit faster (from 107 to 120 fps on doom2 map 1 on my machine), so it may be useful.


38-vita-38:
As you wish.

alexey-lysiuk:
> Please put all code you added into conditionals that only get executed on Linux/Unix. This will cause problems on other platforms.

What kind of problems? `_GNU_SOURCE` is defined for PrBoom's code only when the function was found. Actually, the initial commit builds on all platforms.
I suspect that location of `_GNU_SOURCE` definition can be wrong. If it affects the following checks, then it needs to be moved after them.

coelckers:
Even if it currently compiles, it should be placed in conditionals so that
one can see where it belongs. I don't want to end up with a project where
countless special cases are in but nobody can tell what platform they are
needed for.
There's also no guarantee that it won't cause problems later on.

alexey-lysiuk:
> Even if it currently compiles, it should be placed in conditionals so that one can see where it belongs. I don't want to end up with a project where countless special cases are in but nobody can tell what platform they are needed for.

The `sched_setaffinity` function is either present or not. For proper detection and usage, it requires `_GNU_SOURCE` to be defined before inclusion of `<sched.h>` header file.
The given function is not tied to a particular OS, but to an availability of non-standard POSIX extension.
Cygwin may or may not support it at the moment, but this may change in the future. Same for MinGW. Same for Apple's OSes.

> There's also no guarantee that it won't cause problems later on.

Sorry, but this can apply to pretty much anything.

My point is to do not have any additional conditions if we can avoid them. Launcher for macOS cannot be built reliably on any other platform, so it was put under `if(APPLE)`. Apparently, the case with `sched_setaffinity` is different. The simpler CMake files remain, the easier it will be to work with them.

coelckers:
Even then it should be under "if (NOT MSVC)", because that part is more or
less a given.
I also disagree with "the simpler, the better", when looking at this thing
years later we wouldn't be able to see anymore what relevance it has on
which platform if it is added unconditionally.

38-vita-38:
What about defining _GNU_SOURCE before including sched.h in [SDL/i.main.c](https://github.com/coelckers/prboom-plus/blob/master/prboom2/src/SDL/i_main.c#L50)?

alexey-lysiuk:
According to [related code](https://github.com/coelckers/prboom-plus/blob/9c49cd70bd0d072831983966721820fa4fb4379a/prboom2/src/SDL/i_main.c#L402-L449), `sched_setaffinity()` is supposed to be used on all platforms other than Windows, if it's available of course.
I changed CMake file myself to follow the same logic. 
Thanks for pointing out that detection was flawed.

PR #44, opened 2020-01-11
adaptations for fluidsynth 2.0 (Tom M.)

fabiangreffrath:
This is SVN commit 4543.


coelckers:
I say the same here as for the other sound related things. They are put on hold until I find some time to evaluate the option of replacing the sound system.

fabiangreffrath:
> I say the same here as for the other sound related things. They are put on hold until I find some time to evaluate the option of replacing the sound system.

This was about half a year ago. Is anything going to happen in this regard anytime soon? Because,, if we'll be left with the current sound system for still quite some time, we could just as well commit the fixes to it now.

fabiangreffrath:
With @coelckers announced replacement of the entire sound system not going to happen anytime soon, I'd like to revive this pull request.

PR #45, opened 2020-01-15
unconditionally build with DOGS support

fabiangreffrath:
This fixes the demo desyncs found in

EVIT24-832.lmp  EVIT30-250.lmp  va18-931.lmp  va24-1143.lmp

Has most probably been puzzled up by commit 923bb1b4e which already
unconditionally enabled this in src/info.[ch].

### DIFF
--- a/prboom2/src/doomstat.c
+++ b/prboom2/src/doomstat.c
@@ -88,10 +88,8 @@ int default_monster_infighting=1;
 int monster_friction=1;       // killough 10/98: monsters affected by friction
 int default_monster_friction=1;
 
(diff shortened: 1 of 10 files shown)


fabiangreffrath:
Why would one want to build without dogs nowadays? We even have DFSG compliant replacement for the Sprite and sounds available. Let's just remove this option alltogether.

coelckers:
What's the point of building it without dogs? I never understood why this was conditional, Get rid of the option. PrBoom without dog support sounds like a mostly useless proposition.

PR #49, opened 2020-01-17
Static CRT for MSVC targets

alexey-lysiuk:
Making a PR to confirm that there is no concern with switching to static runtime by default.

Updated dependencies package is [here](https://github.com/coelckers/prboom-plus/releases/download/windows_dependencies/windows_dependencies_v3.zip).


fabiangreffrath:
So this means we get one giant EXE but less DLLs?

alexey-lysiuk:
No, this means no [Microsoft Visual C++ Redistributable for Visual Studio 2015, 2017 and 2019](https://support.microsoft.com/en-us/help/2977003/the-latest-supported-visual-c-downloads). One .exe requires lots of additional efforts though.

coelckers:
This only makes sense if all the DLLs get adjusted as well. Currently they all also depend on the MSVC runtime in DLL form.

alexey-lysiuk:
> This only makes sense if all the DLLs get adjusted as well. Currently they all also depend on the MSVC runtime in DLL form.

[The link](https://github.com/coelckers/prboom-plus/releases/download/windows_dependencies/windows_dependencies_v3.zip) to dependencies package (version 3) I posted above contains all libraries rebuilt with the static runtime. That was the main point of so many adjustments to Windows dependencies workflow.

[Unversioned](https://github.com/coelckers/prboom-plus/releases/download/windows_dependencies/windows_dependencies.zip) package with Windows dependencies wasn’t updated yet. I plan to replace it after merging of this PR.

coelckers:
These are 5 MB larger than the old ones - this really only makes sense if
most dependencies get linked statically.
If you built all these with vcpkg, why can't they be done as static
libraries instead of DLLs? Or do the projects not support this? Most should.

alexey-lysiuk:
SDL2 and Glib (FluidSynth’s dependency) do not support building as static libraries on Windows.
Maybe there are others, but I stopped trying after these failures.

If you want to stick with dynamic runtime, I’m fine with it. Still, I would like to add the static CRT option, just disabled by default.

coelckers:
For SDL there's an official package, and FluidSynth may become redundant when I find time to migrate the sound backend - and a few other dependencies right away with it. Let's not forget that for GZDoom we have a precompiled FluidSynth v1 library. Does v2 provide any significant advantage over v1?

For now I'd rather put this on hold until I can clean up house in the sound system, which seems to require more than half of the needed DLLs. GZDoom has 3 for much better feature support...

kraflab:

rfomin:
Perhaps we should at least update the dependency package.

PR #53, opened 2020-01-18
Add mouse strafe divisor setting

kraflab:
Hello :slightly_smiling_face: 

This PR adds a setting for the mouse strafe divisor (factor by which mouse strafe is reduced as compared to normal horizontal sensitivity), which is particularly important for speedrunners.

When performing certain tricks (e.g., unguided glides), it's helpful to line up using mouse strafe. Unfortunately, mouse strafe is extremely fast by default, making precise movement impossible, except for those with low mouse sensitivity.

A partial mitigation was already in the code base: dividing the strafe by 4. Since this value's relevance depends on a user's sensitivity, I think making it a setting is more natural and increases its usefulness.

### DIFF
--- a/prboom2/src/e6y.c
+++ b/prboom2/src/e6y.c
@@ -517,11 +517,6 @@ dboolean HaveMouseLook(void)
   return (viewpitch != 0);
 }
 
-int GetMouseStrafeDivisor(void)
-{
-  return movement_mousestrafedivisor;
-}
-
 void CheckPitch(signed int *pitch)
 {
   if(*pitch > maxViewPitch)
--- a/prboom2/src/e6y.c
+++ b/prboom2/src/e6y.c
@@ -135,6 +135,7 @@ int movement_shorttics;
 int movement_mouselook;
 int movement_mouseinvert;
 int movement_maxviewpitch;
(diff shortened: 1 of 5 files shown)


fabiangreffrath:
How about two separate sensitivity settings for turning and strafing? I have it like this in Crispy. 

https://github.com/fabiangreffrath/crispy-doom/commit/ff8f4bad5a584841f5db862a04d5893f244e0c67

kraflab:
There's a few reasons:

1) At least with my dpi, the lowest on the bar is still a bit high for strafing. What if a user plays with 1 for turning?
2) The change is backwards compatible in line with the way the code already works.
3) This is more of an "advanced" setting that your average user doesn't need so I didn't feel it belonged on the mouse sensitivity menu.
4) It was easier to implement this way.

I implemented this for myself so it inevitably works the way it was preferable to me. I thought others could benefit from this now that we are kind of having an almost official fork though 🤠

fabiangreffrath:
I see. Anyway, please get rid of the getter function and use the variable directly.

kraflab:
Alright, I made the switch.

fabiangreffrath:
Cool, thanks!

PR #54, opened 2020-01-19
Added Chocolate Doom mouse behaviour option

elim2g:
- Added menu item for Chocolate Doom mouse behaviour
- Chocolate Doom mouse behaviour will carry the result of the mouse delta across multiple tics, such that low sensitivity users will still eventually TL1 or TR1 if they keep pushing their mouse

This feature is often requested by speedrunners as pr+'s original mouse code will simply discard the mouse input if it is too low to perform a TL1 or TR1. This makes it very difficult to get consistent 1 tic turns for things like facing a wall perpendicular for guideless glides, ZeroPresses etc.


fabiangreffrath:
How about "Carry Fractional Tics" or something like this?

coelckers:
> 
> 
> Why don't we just enable this unconditionally? 


Are you sure that all users want this?

elim2g:
> Why don't we just enable this unconditionally? Also, "Chocolate" is misleading as this still isn't identical to Choco's mouse handling code.

While I would love to, there are plenty of runners who have been using old pr+ mouse code for years (eg 4shockblast, Ancalagon) and would likely prefer the option of using the existing code.



> How about "Carry Fractional Tics" or something like this?

Yeah that's fine with me. I don't think new runners will understand what that means, so we could probably open discussion for a more user-friendly name. I believe I actually stole that code from CNDoom, which I figured was unmodified Chocolate but I appear to have mislead myself :D

kraflab:
Maybe "Carry Small Angles"? Whatever it's called we probably will just have to point users to it when they raise the issue, but mentioning small angles might help.

elim2g:
Thanks to Kraf's reminder, I've renamed the mouse_chocolate variable and the in-menu string. I can't think of a better way to describe its purpose clearly to an uninformed user, but that's something our speedrunning Discord can handle manually.

elim2g:
> I'm still a bit puzzled that this quirk needs to be applied at two places in the code when Choco gets away with only one, but that's one for demo recording and one for regular play, I guess. Sigh, what a pleasant code base.

Would you like me to see if there's a clean way around this? I'm not a Doom source guru but I can fiddle around and see what I can do

fabiangreffrath:
> > I'm still a bit puzzled that this quirk needs to be applied at two places in the code when Choco gets away with only one, but that's one for demo recording and one for regular play, I guess. Sigh, what a pleasant code base.
> 
> Would you like me to see if there's a clean way around this? I'm not a Doom source guru but I can fiddle around and see what I can do

Yes, please.

I guess the `if (mouse_carrytics)` code in `G_WriteDemoTiccmd()` is wrong. We should probably leave `cmd->angleturn` as it is in `G_BuildTiccmd()`, get rid of `fudgea()` and add its code to the end of `G_BuildTiccmd()` - just as in Choco. Then, the value of  `cmd->angleturn` will only have to get mangled when actually writing a demo tic in `G_WriteDemoTiccmd()`.

Sorry if I soundconfusing, this is just an idea from looking at the code in a browser window...

kraflab:
@fabiangreffrath @elim2g Please take a look at this:
https://github.com/elim2g/prboom-plus/pull/1/commits/d0099dd6717a6e7c606cb5ac043c5957482eeba2
https://github.com/elim2g/prboom-plus/pull/1

fabiangreffrath:
> @fabiangreffrath @elim2g Please take a look at this:
> [elim2g@d0099dd](https://github.com/elim2g/prboom-plus/commit/d0099dd6717a6e7c606cb5ac043c5957482eeba2)
> [elim2g#1](https://github.com/elim2g/prboom-plus/pull/1)

This is probably what I had in mind, thanks!

elim2g:
I've merged Kraf's PR!

fabiangreffrath:
@coelckers LGTM now, what do you think?

kraflab:
Is there anything else we need here? :slightly_smiling_face:

fabiangreffrath:
LGTM, just waiting for @coelckers 's approval.

fabiangreffrath:
I feel like merging some PRs today.

PR #56, opened 2020-01-24
Add stroller param

kraflab:
You may or may not have heard about a speedrun category that has been gaining some popularity over the past year: stroller (it's now more active than nm 100s...).

The rules are: pacifist + no strafe + turbo 50

This PR adds a simple convenience parameter to the command line that sets turbo 50 and prevents strafing.

If you see this as too frivolous to add, I'll understand :wink:

### DIFF
--- a/prboom2/src/e6y.c
+++ b/prboom2/src/e6y.c
@@ -285,7 +285,7 @@ void e6y_InitCommandLine(void)
     avi_shot_fname = myargv[p + 1];
   stats_level = M_CheckParm("-levelstat");
 
-  if (stroller = M_CheckParm("-stroller"))
+  if ((stroller = M_CheckParm("-stroller")))
   {
     M_AddParam("-turbo");
     M_AddParam("50");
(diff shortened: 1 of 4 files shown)


kraflab:
@coelckers do you have an opinion on this? 😄

coelckers:
I'd like a second opinion first.

prboom2/src/e6y.c
```diff
@@ -285,6 +285,12 @@ void e6y_InitCommandLine(void)
     avi_shot_fname = myargv[p + 1];
   stats_level = M_CheckParm("-levelstat");
 
+  if (stroller = M_CheckParm("-stroller"))
```
fabiangreffrath: Please use double-parentheses if using an assignment inside a check.

kraflab: Done! Been a while since I used C... 🙃

fabiangreffrath:
> I'd like a second opinion first.

Hm, not sure. The same can be achieved by clearing all strafe key bindings and running the port with `-turbo 50` - which is quite a hassle of course. I am just afraid that the "stroller" hype is over before this port sees its next release and then we will have to carry this command line option forever.

kraflab:
Since it's an official category on dsda now, I don't expect it will go anywhere 😄

fabiangreffrath:
Oh, I didn't even know. :+1:

PR #60, opened 2020-02-04
Improve boom random seed

kraflab:
One of the "features" of boom is that the random engine is seeded at the start so you get a different sequence of random numbers (in contrast to vanilla, which is fixed).

For some reason, in prboom+, the random seed was set up using the milliseconds since SDL initialization. As you can imagine, this means you only see a few different seeds.

Using `time(NULL)` isn't going to win any awards for randomness but I think it's satisfactory for this application.

Thanks to vita for bringing this up, because I forgot about it.

### DIFF
--- a/prboom2/src/SDL/i_system.c
+++ b/prboom2/src/SDL/i_system.c
@@ -43,6 +43,7 @@
 #include <stdarg.h>
 #include <stdlib.h>
 #include <ctype.h>
+#include <time.h>
 #include <signal.h>
 #ifdef _MSC_VER
 #define    F_OK    0    /* Check for file existence */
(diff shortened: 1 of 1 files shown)


coelckers:
Is the Boom seed ever being used for demos? If no, this looks fine, if yes, it needs to be optioned.

kraflab:
It is used for demos - that's the main motivation to fix it. The seed is stored in the header so it doesn't affect compatibility. Since the current seed is stuck to a fixed range, and that range depends on the performance of a player's PC, players don't have an equal opportunity on demos. So in that sense this is kind of a bug fix. What would be the motivation for hiding it behind an option (not that I'm outright against that)?

kraflab:
(this is coming out of discussion in the speedrunning discord)

coelckers:
I was unsure if it affects demo sync. Of course when the seed is stored inside the demo that's not a concern.

PR #61, opened 2020-02-06
Fix Boom autoswitch behavior

38-vita-38:
[My DW post](https://www.doomworld.com/forum/post/2061995):
> When you run out of ammo with cl9 or higher, PrBoom+ continuously tries to autoswitch to the most preferred weapon and doesn't let the player switch to what they want. E.g. I spent all my cells with the plasmagun and want to use the rocket launcher but the port inserts a bunch of commands to change to ssg (one can notice this by inspecting Boom or MBF demos in XDRE); I have to switch to it first and only then to RL.

### DIFF
--- a/prboom2/src/g_game.c
+++ b/prboom2/src/g_game.c
@@ -152,6 +152,9 @@ int             totalleveltimes;      // CPhipps - total time for all completed 
 int             longtics;
 int             bytes_per_tic;
 
+dboolean boom_autoswitch;
+dboolean done_autoswitch;
+
 // e6y
 // There is a new command-line switch "-shorttics".
 // This makes it possible to practice routes and tricks
@@ -564,9 +567,14 @@ void G_BuildTiccmd(ticcmd_t* cmd)
   //
   // killough 3/26/98, 4/2/98: fix autoswitch when no weapons are left
 
+  // Make Boom insert only a single weapon change command on autoswitch.
   if ((!demo_compatibility && players[consoleplayer].attackdown && // killough
-       !P_CheckAmmo(&players[consoleplayer])) || gamekeydown[key_weapontoggle])
+       !P_CheckAmmo(&players[consoleplayer])) && !done_autoswitch && boom_autoswitch ||
+       gamekeydown[key_weapontoggle])
+  {
     newweapon = P_SwitchWeapon(&players[consoleplayer]);           // phares
+    done_autoswitch = true;
+  }
   else
     {                                 // phares 02/26/98: Added gamemode checks
       if (next_weapon)
--- a/prboom2/src/g_game.h
+++ b/prboom2/src/g_game.h
@@ -223,6 +223,9 @@ extern int shorttics;
 extern int longtics;
 extern int bytes_per_tic;
 
+extern dboolean boom_autoswitch;
+extern dboolean done_autoswitch;
+
 #define singleplayer (!demorecording && !demoplayback && !democontinue && !netgame)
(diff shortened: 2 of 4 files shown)


fabiangreffrath:
What if you record a (Boom or MBF) demo with the current autoswitch behaviour and then play it back with your patch applied?

38-vita-38:
It will play back just fine.

kraflab:
Thanks for the patch vita 👍
The current code inserts commands on behalf of the player. For demo playback it's reading the commands that were already written so it shouldn't affect compatibility.

kraflab:
Anything else we need here?

kraflab:
Is there already an accessible variable storing the "weapon is switching" status? Maybe that could be used instead of the `done_autoswitch` variable? As in "insert a switch command if we aren't already switching". Is that an option @38-vita-38 ?

38-vita-38:
> Hm, could we possibly get away with less than two new global variables?

> Is there already an accessible variable storing the "weapon is switching" status? Maybe that could be used instead of the `done_autoswitch` variable? As in "insert a switch command if we aren't already switching". Is that an option @38-vita-38 ?

As far as I can remember, no. You can try replacing ``!done_autoswitch`` in the condition with ``players[consoleplayer].pendingweapon == wp_nochange`` but then prb+ won't change weapons sensibly in these examples:
1. your current weapon is bfg, you have 0 cells, try shooting and switching to plasma before or after the switch to ssg takes place, the game will forcefully change to ssg anyway.
2. you have 0 cells; shoot with any weapon and try repeatedly changing to plasma or bfg, that will not work.

I tried a few other combinations of conditions like checking the pendingweapon on the previous tic and nothing worked.

fabiangreffrath:
Hm, I was expecting `player->pendingweapon == wp_nochange` to be a suitable indicator for this...

kraflab:
If the alternative doesn't work in those cases, I think this is a small enough patch to not be an issue. What do you think? :smile:

fabiangreffrath:
> What do you think?

Well, if it helps...

Honestly, I'd rather like to find out why cheking for `player->pendingweapon == wp_nochange` doesn't work as expected, but granted Boom's weapon switching code is a bloody mess already, so applying another layer of band aid shouldn't hurt too much. :wink:

fabiangreffrath:
@coelckers should we merge this?

fabiangreffrath:
Could you guys probably have a look at the `P_SwitchWeapon()` implementation in MBF before we merge this? It is supposed to fix "BFG/SSG bugs".

https://github.com/fabiangreffrath/woof/blob/f2e4320419527146af5d4e9bc2f4bb16d3bf0014/Source/p_pspr.c#L157

kraflab:
What exactly are we looking for? This method is identical to the implementation already in prboom+ as far as I can see.

fabiangreffrath:
Yes, sorry. This was just an idee-fix and I had to type it on the phone in order not to forget it.

I have just read this comment in the MBF source code and thought, maybe the underlying issue is in `P_SwitchWeapon()` and we don't have to introduce new matters to break the loop. Obviously, this wasn't the case, sorry for the noise.

kraflab:
No worries :ok_hand:

kraflab:
Well, merge or no? :smile:

fabiangreffrath:
If someone could point out to me why this doesn't work by checking for `player->pendingweapon == wp_nochange` instead of the newly introduced `done_autoswitch` I'd be more convinced.

PR #63, opened 2020-02-15
Add configurable quickstart window

kraflab:
## Background
If you perform some inputs during the melt when you launch doom, then they will "stack up" and execute after the melt instantly - e.g., turn and straferun during melt will become one frame of turning + straferunning followed by a series of straferunning frames before the player has control again. The details of this action depend on the source port. However, the first frame will be empty, because of how input is read before the melt starts. However, you _can_ get first frame movement if your timing is precise enough. This is called getting a "quickstart" in the speedrunning community. 

The problem is that how precise you need to be depends on both hardware and the source port. Thus, some users are at a disadvantage relative to others. For most players this isn't relevant, but for the players pushing the human limit, one frame makes all the difference (e1m1 in 8.97 comes to mind, and doom 2 map 1 -nomonsters is at 5.03 right now).

## Comparison
Right now crispy doom is the "easy" port for this. You can tell you execute quickstart when it shows a different angle during the melt than the default one for a given map, and the way to execute it is quite consistent. With current prboom+ I can execute a quickstart, but the window is extremely tight. 4shockblast can't get it at all, but he could on an older computer just fine. The point is that this isn't a skill question, but rather a question of hardware and software.

## Goal
The goal of this PR is to take out the hardware / software dependent element in executing a quickstart, so that every player is on an even playing field.

## Solution
I've added an option to the settings that simply adds a delay to the first gametic before reading the input. This simulates a slower startup and only affects the game during demo recording.

### DIFF
--- a/prboom2/src/d_client.c
+++ b/prboom2/src/d_client.c
@@ -336,6 +336,9 @@ void NetUpdate(void)
     lastmadetic += newtics;
     if (ffmap) newtics++;
     while (newtics--) {
+      if (quickstart_window_ms > 0 && gametic == 0 && demorecording)
+        I_uSleep(quickstart_window_ms * 1000);
+
       I_StartTic();
       if (maketic - gametic > BACKUPTICS/2) break;
       
--- a/prboom2/src/d_client.c
+++ b/prboom2/src/d_client.c
@@ -336,9 +336,6 @@ void NetUpdate(void)
     lastmadetic += newtics;
     if (ffmap) newtics++;
     while (newtics--) {
-      if (quickstart_window_ms > 0 && gametic == 0 && demorecording)
-        I_uSleep(quickstart_window_ms * 1000);
-
       I_StartTic();
       if (maketic - gametic > BACKUPTICS/2) break;
       
--- a/prboom2/src/d_main.c
+++ b/prboom2/src/d_main.c
@@ -446,6 +446,9 @@ static const char *auto_shot_fname;
 
 static void D_DoomLoop(void)
 {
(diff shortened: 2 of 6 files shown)


kraflab:
@fabiangreffrath what do you think? This has a green light from looper and shockblast by the way 😉

fabiangreffrath:
If it works, I am all for it!

However, I have not checked myself if it works as intended. I just know that CnDoom had a similar artificial delay implemented once and reportedly it was not as efficient as it promised.

Regarding the placement of the code. AFAIUI, this delay is to be introduced only once - at gametic 0. However, the code is injected into `NetUpdate()` which is called during each tic, and only if we are building with `HAVE_NET` defined. 

Wouldn't it be better inserted into `D_DoomLoop()` right before the `for()` loop?

kraflab:
@fabiangreffrath I've moved the code as you suggested, that does indeed make more sense :smile: 

This works perfectly for me at least (frame zero turn + sr40):
```
Frame Run Strafe Turn Extra
    0   50  -40    3 | -
    1   50  -40    0 | -
    2   50  -40    0 | -
    3   50  -40    0 | -
    4   50  -40    0 | -
```
Since the duration is configurable, it hopefully works for everyone.

kraflab:
Shall we merge? :slightly_smiling_face:

fabiangreffrath:
Could you please remove the `demorecording` condition? It should be possible to practice with this setting without needing to actually record a demo each time.

kraflab:
Done.

fabiangreffrath:
Merged, thanks!

PR #75, opened 2020-04-11
Show Extended Help screen when pressing the Help key.

JadingTsunami:
This PR proposes a fix to a very old bug that broke the Extended Help functionality of Boom.

After the final Boom 2.02 release, around 10/1998 the Help menu accessed by the F1 hotkey was refactored to a single menu definition instead of calling versions of the "Read This!" menu item.

However, the code that installs the Extended Help handler was not updated to redirect calls to the new Help menu. This effectively made it impossible to use Extended Help without a "Read This!" menu item.

This PR adds the missing call to install the handler and restores the F1 hotkey functionality as it was in the Boom releases.

### DIFF
--- a/prboom2/src/m_menu.c
+++ b/prboom2/src/m_menu.c
@@ -4054,7 +4054,15 @@ void M_InitExtendedHelp(void)
     i = W_CheckNumForName(namebfr);
     if (i == -1) {
       if (extended_help_count) {
-        /* Update consolidated help menu */
+        /* The Extended Help menu is accessed using the
+         * Help hotkey (F1) or the "Read This!" menu item.
+         *
+         * If Extended Help screens are present, use the
+         * Extended Help routine when either the F1 Help Menu
+         * or the "Read This!" menu items are accessed.
+         *
+         * See also: https://www.doomworld.com/forum/topic/111465-boom-extended-help-screens-an-undocumented-feature/
+         */
           HelpMenu[0].routine = M_ExtHelp;
         if (gamemode == commercial) {
           ExtHelpDef.prevMenu  = &ReadDef1; /* previous menu */
--- a/prboom2/src/m_menu.c
+++ b/prboom2/src/m_menu.c
@@ -4054,13 +4054,15 @@ void M_InitExtendedHelp(void)
     i = W_CheckNumForName(namebfr);
     if (i == -1) {
       if (extended_help_count) {
+        /* Update consolidated help menu */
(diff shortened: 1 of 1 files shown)


fabiangreffrath:
Sorry, but I fail to understand what this is supposed to fix. Could you probably provide an example what happens now and what changes with your patch applied?

JadingTsunami:
> Sorry, but I fail to understand what this is supposed to fix. Could you probably provide an example what happens now and what changes with your patch applied?

Certainly; [linked here](https://www.doomworld.com/forum/topic/111465-boom-extended-help-screens-an-undocumented-feature/) is a thread on the feature. In the last post there is a detailed breakdown of the original (intended) behavior and bug history.

I submitted a PR because I received some requests to help restore the functionality. Since I had already done all the research and identified the fix, it seemed reasonable to consolidate it here.

fabiangreffrath:
So, this fix only becomes apparent for WADs that include `HELPnn` named help screen lumps? And it only applies to Doom 2 because this doesn't have a "Read Me!" menu entry while Doom 1 has, right?

JadingTsunami:
> So, this fix only becomes apparent for WADs that include `HELPnn` named help screen lumps? 

Yes, exactly right.

> And it only applies to Doom 2 because this doesn't have a "Read Me!" menu entry while Doom 1 has, right?

It applies to both Doom and Doom 2 in that it fixes the F1 key's functionality when `HELPnn` lumps are present. This bug didn't affect the "Read This!" menu, which is why Doom still had a route to view the `HELPnn` lumps when they were present. But that was not the only intended way.

fabiangreffrath:
I see. LGTM then.

JadingTsunami:
> One request though. could you elaborate a bit in the comment? This one sentence doesn't really do justice to the complexity of the issue.

Certainly! I have pushed a commit with an updated comment. If you have any feedback let me know.

fabiangreffrath:
@coelckers It is currently unclear to me how we should proceed with the open and already reviewed pull requests, i.e. this one and #54. I am hesitating to merge them myself, because I'd like to hear your final opinion about them first. On the other hand, I have already formally expressed my opinion by checking the "approved" button, which to me means "merge if you like". Would you like me to merge them myself or do you just need more time to thoroughly review them yourself?

JadingTsunami:
Are there any lingering concerns about this PR? I'd be happy to address them if so.

fabiangreffrath:
I have taken the liberty, thanks!

PR #77, opened 2020-04-22
Bug fixes

Ritchie333:
Two bug fixes found testing latest branch on Debian 10 with Ultimate Doom IWAD

1. SDL sound can be noisy, clearing the buffer fixes this
2. After completing any first map on any episode, play went on to E1M2 regardless of episode.


.gitignore
```diff
@@ -4,3 +4,20 @@ prboom2/data/prboom.wad
...
+data/Makefile
+data/cmake_install.cmake
+prboom2/data/rdatawad
+glboom-plus
+prboom-plus
+prboom-plus-game-server
+prboom-plus.wad
+src/CMakeFiles
+src/Makefile
+src/cmake_install.cmake
```
alexey-lysiuk, line 23: There is no need to add all this. Please use designated build directories named `build*` instead. While doing in-tree build isn't prohibited, it's not recommended either.

prboom2/src/g_game.c
```diff
@@ -1586,6 +1586,12 @@ void G_DoCompleted (void)
 		  goto frommapinfo;	// skip past the default setup.
 	  }
   }
+  else
```
fabiangreffrath, line 1589: Wasn't this already fixed by https://github.com/coelckers/prboom-plus/commit/dfe091e35a96e014d3f77c23b198f4a5004c6004 ?

Ritchie333: Looks like it given the date (6 Sep 2019) - the fix I did was some time before that but I only just thought to review changes now.

.gitignore
Ritchie333: Should the documentation not say something about this? The INSTALL file still talks about using GNU autoconf.

prboom2/src/g_game.c
fabiangreffrath: When you said "latest branch" I thought you meant that.

.gitignore
alexey-lysiuk: If you can update it, please do ;)

fabiangreffrath:
@Ritchie333 could you possibly post the sound buffer clearing fix as a separate PR, please?

fabiangreffrath:
> @Ritchie333 could you possibly post the sound buffer clearing fix as a separate PR, please?

Commited, thanks!

PR #79, opened 2020-05-02
Make armor color depend on type, not amount

physixcat:
I have edited both the Status Bar and the Advanced HUD so that the armor color only depends on the last type of armor that was picked up, and never on the percentage amount. This means that blue armor will display as blue even if it dips below 100%, and green armor will display as green even if it raises above 100%. If any armor reaches 0%, it will display as red.

I think this is a quality of life change because it reminds players of the last armor type they picked up, so they may more easily decide when to switch to another one during play. This is in line with the behavior displayed by Crispy Doom's armor color display.


prboom2/src/hu_stuff.c
```diff
@@ -1249,6 +1249,16 @@ int HU_GetHealthColor(int health, int def)
 
 int HU_GetArmorColor(int armor, int def)
 {
+// [pull req] suggested change: armor color dictated by type (Advanced HUD)
```
fabiangreffrath, line 1252: Once we merge your PR, these comments will not make any sense anymore. Please remove phrases like "pull req" and "suggested changes".

prboom2/src/hu_stuff.c
```diff
@@ -1249,6 +1249,16 @@ int HU_GetHealthColor(int health, int def)
...
+  if (plr->armortype >= 2)
+      return CR_BLUE;
+  else if (plr->armortype == 1)
+      return CR_GREEN;
+  else if (plr->armortype == 0)
+      return CR_RED;
+
+// [old behavior] armor color dictated by percent only, not type (Advanced HUD)
```
fabiangreffrath, line 1260: I'd rather prefer if there was a switch to choose between "old" (i.e. by value) and "new" (i.e. by type) behaviour in the HUD menu. Not because I like menu switches so much, but because many players will have got used to the old behaviour and will get confused if this changes suddenly. Also, all the menu items concerning the armor color thresholds in the HUD menus would suddenly have no meaning anymore.

prboom2/src/hu_stuff.c
physixcat: Understood! Apologies for including those phrases in the comments, they were basically just for myself. They have been removed in the latest commit.

prboom2/src/hu_stuff.c
physixcat: Thank you for the feedback! I have implemented the new behavior as a toggle in the HUD menu in this latest commit. It is right above the armor color thresholds.

physixcat:
I have made the requested changes in the latest commit. Is the current implementation satisfactory?

physixcat:
Sure! Is this latest commit what you had in mind?

fabiangreffrath:
Top, thanks!

PR #83, opened 2020-05-19
Temporary fix for SDL_Audio output stream buffer clear bug

Gustavo6046:
I have not yet figured out why setting the buffer to zero would break the MIDI player, since SDL_Audio passes the argument as a pointer to its stream buffer. Perhaps it conflicts with the MIDI player somehow; perhaps SDL's MIDI player writes to the same buffer _before_ our callback?

Either way, I have put a temporary fix, which simply skips the buffer memset if the MIDI player is not the SDL one.


Gustavo6046:
Done.

Gustavo6046:
This time, I double checked with `git diff 9802889..` - there are absolutely no changes anymore, other than the three important lines here.

Gustavo6046:
There has been lots of work in the `gus-exeprimental` branch. It does not concern PRBoom+ particularly; it is actually a build with a bot (work in progress) and where -solo-net overrides no_net in m_cheat.c. For that reason I put it into a separate branch. However, it may be worth a look. It may have a minor change or two worth of being pulled into PRBoom+ (like the solo-net tweak), but I am not sure.

fabiangreffrath:
Sorry, but nobody is probably going to review your entire branch. If there are changes that you consider applicable to the master branch, please file PRs for them. Thanks.

Gustavo6046:
It is fine. The point is that it is possible that, in the near future, I may have more than just these three petty lines to contribute.

fabiangreffrath:
Looking forward to it. :+1:

Gustavo6046:
They will be in separate PRs, though. I won't add any changes to this one – other than any requested ones, of course.

Gustavo6046:
Oh, regarding your last comment – didn't see that one – thank you very much! It is good to know that this simple solution is consistent enough.

In fact, maybe this is more than a temporary solution. I suggest clearing the buffer before (or at the begin of) the routine that plays MIDI through SDL (or whatever you use in the SDL MIDI player option); then, in any other case, be pragmatic, and clear the buffer in the start of the regular SDL_audio callback, instead.

fabiangreffrath:
> They will be in separate PRs, though. I won't add any changes to this one – other than any requested ones, of course.

Of course, that's the expected workflow. :wink:

Issue #84, opened 2020-05-20
prboom-plus: Heap buffer overflow in UDP code (CVE-2019-20797)


fabiangreffrath:
This was reported yesterday against the prboom-plus Debian package (https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=961031). There is already a patch available for the first issue (251 on  Sourceforge). For the latter two it makes probably sense to limit the number of tics which are meant to be sent over network to a reasonable amount, e.g. 35 (i.e. a second worth of gameplay, such a lag would make the network game unplayable anyway).

---

Dear Maintainer,

Description:
"An issue was discovered in e6y prboom-plus 2.5.1.5. There is a buffer
overflow in client and server code responsible for handling received UDP
packets, as demonstrated by I_SendPacket or I_SendPacketTo in
i_network.c."

URLs:
* https://logicaltrust.net/blog/2019/10/prboom1.html
* https://sourceforge.net/p/prboom-plus/bugs/251/
* https://sourceforge.net/p/prboom-plus/bugs/252/
* https://sourceforge.net/p/prboom-plus/bugs/253/

PR #85, opened 2020-05-20
fix heap buffer overflows in UDP code (CVE-2019-20797)

fabiangreffrath:
* Limit length of buffer passed over to ChecksumPacket().
  Patch taken from the OP at https://logicaltrust.net/blog/2019/10/prboom1.html

* Never send more than one second worth of tics (i.e. 35) in both the
  main() routine in d_server.c and NetUpdate() in d_client.c.
  This avoids overflows of the allocated UDF buffer with a fixed size
  of 10000 bytes. Theoretically, up to about 35 seconds could be sent in the
  client code and up to about 7 seconds in the server code, but the
  network game would be unplayable with such a lag anyway.

  Client code: pkt_size = 8 + 2 + X * 8 (= 9810 for X = 35*35)
  Server code: pkt_size = 8 + 1 + X * (1 + 4 * (1 + 8)) (= 9074 for X = 7 * 35)

Fixes: #84

### DIFF
--- a/prboom2/src/SDL/i_network.c
+++ b/prboom2/src/SDL/i_network.c
@@ -240,7 +240,7 @@ size_t I_GetPacket(packet_header_t* buffer, size_t buflen)
   checksum=buffer->checksum;
   buffer->checksum=0;
   if ( (status!=0) && (len>0)) {
-    byte psum = ChecksumPacket(buffer, udp_packet->len);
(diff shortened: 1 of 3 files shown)


fabiangreffrath:
Chocolate Doom backs up 128 tics, which is about 2.5 seconds of time.

fabiangreffrath:
Just got confirmation from LogicalTrust that my patch fixes the issue.

Issue #70, opened 2020-03-03
Add install target


alexey-lysiuk:
Transition to CMake removed ability to do `make install`.

fabiangreffrath:
Is this going to change anytime soon?

alexey-lysiuk:
Not from my side.

PR #93, opened 2020-06-06
Add CMake install targets

vilhelmgray:
Fixes: https://github.com/coelckers/prboom-plus/issues/70

### DIFF
--- a/prboom2/data/CMakeLists.txt
+++ b/prboom2/data/CMakeLists.txt
@@ -362,3 +362,5 @@ add_custom_command(
     DEPENDS rdatawad ${WAD_SRC}
 )
 add_custom_target(prboomwad DEPENDS ${WAD_DATA_PATH})
+set(WAD_INSTALL_PATH ${DOOMWADDIR} CACHE PATH "Path to install prboom-plus.wad")
+install(FILES ${WAD_DATA_PATH} DESTINATION ${WAD_INSTALL_PATH} COMPONENT "PrBoom-Plus internal WAD")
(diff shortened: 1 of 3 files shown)


vilhelmgray:
I created a `WAD_INSTALL_PATH` setting so users can choose an install path that is different from `DOOMWADDIR`. Is there any situation where a user might want to do this?

If not, then I can remove this new setting and have **prboom-plus.wad** always install into `DOOMWADDIR`.

fabiangreffrath:
Thanks for doing this! Please keep `WAD_INSTALL_PATH` as a separate setting, it will make life easier for Linux packagers.

prboom2/data/CMakeLists.txt
```diff
@@ -362,3 +362,5 @@ add_custom_command(
     DEPENDS rdatawad ${WAD_SRC}
 )
 add_custom_target(prboomwad DEPENDS ${WAD_DATA_PATH})
+set(WAD_INSTALL_PATH ${DOOMWADDIR} CACHE PATH "Path to install prboom.wad")
```
fabiangreffrath, line 365: It's `prboom-plus.wad`, inherited mistake.

vilhelmgray: I've made the change now to `prboom-plus.wad`. By the way, do you think it would be better for this to say "Path to install the PrBoom-Plus WAD" instead; or is the current text fine?

fabiangreffrath: I guess it's okay as it is, please push that change.

fabiangreffrath:
Top, thanks!

PR #97, opened 2020-06-13
Track demo joins

kraflab:
This PR stores demo joins as a new special input in the 7th bit of the "buttons" byte. The goal here is to track where segments meet for posterity and to assist with automated demo analysis performed for DSDA.

While this bit was unused in the original lmp format, it is sometimes used as an extra bit for weapon swap extensions (8 = chainsaw, 9 = ssg). This is the case for certain old versions of prboom+ (now achieved with the `-emulate` parameter or a specific flag) and for newer compatibility levels (boom, mbf, etc). In either of those contexts, and when a weapon swap is triggered, this bit must be left alone.

While this limitation is unfortunate, it's satisfactory for us to cover the cases we can, and there's only so much you can do without breaking compatibility entirely.

FYI @fabiangreffrath

### DIFF
--- a/prboom2/src/d_event.h
+++ b/prboom2/src/d_event.h
@@ -114,6 +114,8 @@ BT_WEAPONMASK_OLD   = (8+16+32),//e6y
   BTS_SAVEMASK    = (4+8+16),
   BTS_SAVESHIFT   = 2,
 
+  // Demo joined.
+  BT_JOIN = 64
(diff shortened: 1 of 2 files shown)


fabiangreffrath:
Any idea for a pendant commit in Crispy yet?

kraflab:
I should have time to dive into crispy today.

Issue #98, opened 2020-06-16
Umapinfo intermission screens showing wrong episode map in Doom 1


UnluckySpade7:
I've been experimenting with a personal wad that allows me to play all the Doom 1 episodes in a continious playthrough but there's a minor issue exclusive to prboom plus that gzdoom doesn't have related to the map screens.

When I IDCLEV to a map with Umapinfo enabled and finish it the intermission map of the previous episode displays.
![umapinfobug](https://user-images.githubusercontent.com/39349626/84772852-6f63e780-afd3-11ea-9d2a-ba4395463a32.png)
Then when i go to start the next map the regular map shows up but with leftovers from the first.
![umapinfobug2](https://user-images.githubusercontent.com/39349626/84772975-a1754980-afd3-11ea-9012-31948e3c8dad.png)

This also applies to going from episode 2 to episode 3 where the map doesn't get cleared of blood spots from the previous map and the "you are here" arrow points to the wrong part of the map.
![umapinfobug3](https://user-images.githubusercontent.com/39349626/84773265-1183cf80-afd4-11ea-9578-f2ee0f4556a7.png)

Here is the wad I've been using to test this: 
[umaptest.zip](https://github.com/coelckers/prboom-plus/files/4786539/umaptest.zip)

fabiangreffrath:
Does it work if you move the `wminfo.nextep = wminfo.epsd = gameepisode -1;` and `wminfo.last = gamemap -1;` lines from L1615 up to L1569 right before the `if (gamemapinfo)` block?

https://github.com/coelckers/prboom-plus/blob/HEAD/prboom2/src/g_game.c#L1615

UnluckySpade7:
I don't know how I could test that since I've been using pre-made builds from the forum and don't know how to build it myself

fabiangreffrath:
Sorry, I was assuming everybody is a coder. 😁 

The first chunk fixes the intermissions screens during episode switch for me. The second chunk fixes the crash when switching to episode 4. It would be nice if somebody else could confirm these fixes before I apply them.

```
diff --git a/prboom2/src/g_game.c b/prboom2/src/g_game.c
index 5f5e50d2..c809d891 100644
--- a/prboom2/src/g_game.c
+++ b/prboom2/src/g_game.c
@@ -1564,6 +1564,9 @@ void G_DoCompleted (void)
   if (automapmode & am_active)
     AM_Stop();

[...]

PR #99, opened 2020-06-18
UMAPINFO: fix some episode transition issues

fabiangreffrath:
* Fix wminfo.epsd and wminfo.last not being set for UMAPINFO defined
  maps. This would lead to the wrong episode background being drawn on
  the "Finished" intermission screen.
* Reset the didsecret property for all players when switching
  episodes, so the blood splat will not be drawn on ExM9.
* Next episode becomes current episode when switching from the
  "Finished" to the "Entering" intermission screen, so the correct
  background, animations, blood splat and "You are here" pointers are
  used.

Fixes: #98, thanks @UnluckySpade7.

### DIFF
--- a/prboom2/src/g_game.c
+++ b/prboom2/src/g_game.c
@@ -1564,6 +1564,9 @@ void G_DoCompleted (void)
   if (automapmode & am_active)
     AM_Stop();
 
+  wminfo.nextep = wminfo.epsd = gameepisode -1;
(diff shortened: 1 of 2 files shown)


fabiangreffrath:
@coelckers As this affects the UMAPINFO implementation, I'd be glad to get some feedback from you before I merge this.

fabiangreffrath:
More than one week and not even a shrug...

PR #104, opened 2020-07-10
Add "Dropped Item" support

Ferk:
Support the same dehacked Thing field as Doom Retro: https://github.com/bradharding/doomretro/commit/096f5ef09556dfa1cc5f0ab72d7d57f9c8e37010

"Dropped item" will allow to specify the Thing Id to be spawned after the Thing dies.

This was briefly discussed in Doomworld, regarding the extension of dehacked:
  https://www.doomworld.com/forum/topic/115638-dehextra-discussion-split-from-pr-umapinfo-thread/?tab=comments#comment-2155933

### DIFF
--- a/prboom2/src/p_inter.c
+++ b/prboom2/src/p_inter.c
@@ -805,24 +805,10 @@ static void P_KillMobj(mobj_t *source, mobj_t *target)
   // This determines the kind of object spawned
   // during the death frame of a thing.
 
-  switch (target->type)
-    {
-    case MT_WOLFSS:
-    case MT_POSSESSED:
-      item = MT_CLIP;
-      break;
-
-    case MT_SHOTGUY:
-      item = MT_SHOTGUN;
-      break;
-
-    case MT_CHAINGUY:
-      item = MT_CHAINGUN;
-      break;
-
-    default:
-      return;
-    }
+  if (target->info->droppeditem != MT_NULL)
+  {
+    item = target->info->droppeditem;
+  }
 
   mo = P_SpawnMobj (target->x,target->y,ONFLOORZ, item);
   mo->flags |= MF_DROPPED;    // special versions of items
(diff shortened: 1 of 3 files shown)


prboom2/src/p_inter.c
```diff
@@ -804,28 +802,12 @@ static void P_KillMobj(mobj_t *source, mobj_t *target)
...
-  mo->flags |= MF_DROPPED;    // special versions of items
+  if (target->info->droppeditem != MT_NULL)
+  {
+    mobj_t     *mo;
+    mo = P_SpawnMobj (target->x,target->y,ONFLOORZ, target->info->droppeditem);
+    mo->flags |= MF_DROPPED;    // special versions of items
+  }
 
 #ifdef GL_DOOM
   if (target->momx == 0 && target->momy == 0)
```
Ferk, line 813: I'm unsure what the MF_FOREGROUND flag is doing here for GLBOOM, do we only want it in corpses that drop stuff?

Ferk: Well, I'll just leave it with the same behavior as before.

Ferk:
Now Crispy Doom also supports this: https://github.com/fabiangreffrath/crispy-doom/pull/635

prboom2/src/p_setup.c
```diff
@@ -2849,12 +2849,43 @@ void P_SetupLevel(int episode, int map, int playermask, skill_t skill)
   R_SmoothPlaying_Reset(NULL);
 }
 
+//
+// P_InitThingProperties
+//
+// Initialize Thing extra properties (keeping vanilla props in info.c)
+//
+static void P_InitThingProperties (void)
```
fabiangreffrath, line 2857: This loop needs to be moved to `D_BuildBEXTables()`.

Ferk: Thanks

PR #107, opened 2020-07-12
Extensible demo format + "PR+UM" signature

XaserAcheron:
Hey y'all,

There's been some talk about extending the demo format + signature in the Doomworld thread recently, so I figured for convenience's sake, I'd make a PR for anotak's implementation of this. It's gotten some eyes on it in the past, code-review-wise but we may as well get the ball rolling on getting it submitted for real.

More comments on this (plus a quote from Graf showing support for the idea, in case that was in question) on DW [here](https://www.doomworld.com/forum/topic/106866-prboomumapinfo-v-2517/?page=14&tab=comments#comment-2156841).

### DIFF
--- a/prboom2/src/g_game.c
+++ b/prboom2/src/g_game.c
@@ -87,6 +87,9 @@
 #include "e6y.h"//e6y
 #include "statdump.h"
 
+// ano - used for version 255+ demos, like EE or MBF
+static char     prdemosig[] = "PRB+";
+
 #define SAVEGAMESIZE  0x20000
 #define SAVESTRINGSIZE  24
 
@@ -3179,13 +3182,71 @@ void G_BeginRecording (void)
 {
   int i;
   byte *demostart, *demo_p;
+  int num_requirements = 0;
   demostart = demo_p = malloc(1000);
   longtics = 0;
 
-  if (umapinfo_loaded)
+  // ano - jun2019 - add the extension format if needed
+  if (gamemapinfo)
   {
-	  *demo_p++ = 255;
+    num_requirements++;
   }
+
+  if (num_requirements > 0)
+  {
+    // demover
+    *demo_p++ = 0xFF;
+    // signature
+    *demo_p++ = 'P';
+    *demo_p++ = 'R';
+    *demo_p++ = 'B';
+    *demo_p++ = '+';
+    // extension version
+    *demo_p++ = 1;
+    //
+    *demo_p++ = num_requirements;
+
+    if (gamemapinfo)
+    {
+      int mapname_len = strnlen(gamemapinfo->mapname, 9);
+
+      // ano - note that the format has each length by each string
+      // as opposed to a table of lengths
+      *demo_p++ = 0x08;
+      *demo_p++ = 'U';
+      *demo_p++ = 'M';
+      *demo_p++ = 'A';
+      *demo_p++ = 'P';
+      *demo_p++ = 'I';
+      *demo_p++ = 'N';
+      *demo_p++ = 'F';
+      *demo_p++ = 'O';
+
+      // ano - to properly extend this to support other requirement strings
+      // we wouldn't just plop this here, but right now we only support the 1
+      // in the future, we should assume that chunks in the header should
+      // follow the order of their appearance in the requirements table.
+      if (mapname_len > 8)
+      {
+        I_Error("Unable to save map lump name %s, too large.", gamemapinfo->mapname);
+      }
+
+      for (i = 0; i < mapname_len; i++)
+      {
+        // FIXME - the toupper is a hacky workaround for the case insensitivity
+        // in the current UMAPINFO reader. lump names should probably not be
+        // lowercase ever (?)
+        *demo_p++ = toupper(gamemapinfo->mapname[i]);
+      }
+
+      // lets pad out any spare chars if the length was too short
+      for (; i < 8; i++)
+      {
+        *demo_p++ = 0;
+      }
+    }
+  }
+  // ano - done with the extension format!
 
   /* cph - 3 demo record formats supported: MBF+, BOOM, and Doom v1.9 */
   if (mbf_features) {
@@ -3490,7 +3551,9 @@ const byte* G_ReadDemoHeader(const byte *demo_p, size_t size)
 const byte* G_ReadDemoHeaderEx(const byte *demo_p, size_t size, unsigned int params)
 {
   skill_t skill;
-  int i, episode, map;
+  int i, episode, map, extension_version;
+
+  int using_umapinfo = 0;
 
   // e6y
   // The local variable should be used instead of demobuffer,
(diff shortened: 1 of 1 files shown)


prboom2/src/g_game.c
```diff
@@ -3179,13 +3182,71 @@ void G_BeginRecording (void)
...
+  if (num_requirements > 0)
   {
-	  *demo_p++ = 255;
+    // demover
+    *demo_p++ = 0xFF;
+    // signature
+    *demo_p++ = 'P';
+    *demo_p++ = 'R';
+    *demo_p++ = 'B';
+    *demo_p++ = '+';
```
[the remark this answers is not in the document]
Ferk: If Eternity is considering also support for these demos (if UMAPINFO support lands on Eternity), it might also be advisable to agree with them on adding another '\0' as separator after the version number, since Eternity uses 4 bytes for the version.

prboom2/src/g_game.c
```diff
@@ -3512,21 +3575,175 @@ const byte* G_ReadDemoHeaderEx(const byte *demo_p, size_t size, unsigned int par
...
+    // num_requirements *
+    //    1 byte string length
+    //    and length chars (up to 255 obviously)
+    // note that the format has each length by each string
+    // as opposed to a table of lengths
+
+    // an example requirements string is "UMAPINFO".
+    // if num_requirements needs extending past 255 one day,
+    // consider reserving requirement string "MORE", which
+    // would could be implemented to load another list of 255.
```
Ferk, line 3601: This might not ever be needed... but perhaps a better idea than adding a "MORE" string if the count goes past than 255 would be to add the next byte whenever a 255 is found ( so ``xFF x00`` would be 255 elements, ``xFF xFF 40" would be 550 elements, etc.)

prboom2/src/g_game.c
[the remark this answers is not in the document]
Ferk: If Eternity is considering also support for these demos (if UMAPINFO support lands on Eternity), it might also be advisable to agree with them on adding another '\0' as separator after the version number, since Eternity uses 4 bytes for the version.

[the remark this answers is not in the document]
Ferk: If Eternity is considering also support for these demos (if UMAPINFO support lands on Eternity), it might also be advisable to agree with them on adding another '\0' as separator after the version number, since Eternity uses 4 bytes for the version.

prboom2/src/g_game.c
```diff
@@ -3179,13 +3182,71 @@ void G_BeginRecording (void)
 {
   int i;
   byte *demostart, *demo_p;
+  int num_requirements = 0;
```
kraflab, line 3185: The name of this variable is a bit confusing to me. Maybe it should be something like "num_extensions"?

prboom2/src/g_game.c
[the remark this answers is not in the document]
Altazimuth: Ahh crap I think that was my fault from way back when, providing faulty info. Also yes I do plan to bring playback support into EE, as well as recording of cl2.

XaserAcheron:
Just merged in some changes from Altaz that should address all the concerns -- apologies for the delay; GitHub chose not to notify me about this for some reason. o_O

prboom2/src/g_game.c
```diff
@@ -3179,13 +3182,74 @@ void G_BeginRecording (void)
...
-	  *demo_p++ = 255;
+    // demover
+    *demo_p++ = 0xFF;
+    // signature
+    *demo_p++ = prdemosig[0]; // 'P'
+    *demo_p++ = prdemosig[1]; // 'R'
+    *demo_p++ = prdemosig[2]; // '+'
+    *demo_p++ = prdemosig[3]; // 'U'
+    *demo_p++ = prdemosig[4]; // 'M'
+    *demo_p++ = '\0';
```
fabiangreffrath, line 3205: Six bytes for a signature? How many ports are we expecting to support this? 
Also, why do we need a terminating NULL byte? This isn't a proper string, anyway (you don't even use `strcat()` et al here), but just a couple of ASCII bytes embedded into some binary cruft. I'd prefer if the signature would be four bytes, e.g. "PR+U".

prboom2/src/g_game.c
```diff
@@ -3179,13 +3182,74 @@ void G_BeginRecording (void)
...
+      // as opposed to a table of lengths
+      *demo_p++ = 0x08;
+      *demo_p++ = 'U';
+      *demo_p++ = 'M';
+      *demo_p++ = 'A';
+      *demo_p++ = 'P';
+      *demo_p++ = 'I';
+      *demo_p++ = 'N';
+      *demo_p++ = 'F';
+      *demo_p++ = 'O';
```
fabiangreffrath, line 3226: Why do we have to embed the entire "UMAPINFO" string here? And whats the matter with `0x08`? Can't we just enumerate all possible features and assign them simple numerical values?

prboom2/src/g_game.c
```diff
@@ -3179,13 +3182,74 @@ void G_BeginRecording (void)
...
+      {
+        I_Error("Unable to save map lump name %s, too large.", gamemapinfo->mapname);
+      }
+
+      for (i = 0; i < mapname_len; i++)
+      {
+        // FIXME - the toupper is a hacky workaround for the case insensitivity
+        // in the current UMAPINFO reader. lump names should probably not be
+        // lowercase ever (?)
+        *demo_p++ = toupper(gamemapinfo->mapname[i]);
```
fabiangreffrath, line 3242: Why do we store the map lump name here? It's also stored in the regular demo header and at this point it makes no difference if the level is loaded by means of UMAPINFO or not - it's just the level progression that matters. To make sure that this gets right, we need to enforce the UMAPINFO lump getting loaded - which is the entire purpose of this new header.

prboom2/src/g_game.c
Altazimuth: This, Eternity, and MBF all do this. Six bytes. The intent is that EE can read these too.

prboom2/src/g_game.c
Altazimuth: 100% certainty there will be no clashes with anything else? 8 bytes a demo isnt going to cause any real issues, is it?

prboom2/src/g_game.c
fabiangreffrath: Ah, indeed, even MBF has such a weird signature in the header.

prboom2/src/g_game.c
fabiangreffrath: Ah, OK, so it's meant to indicate the length of the identifier string that follows. I didn't get that on first parsing.

prboom2/src/g_game.c
Ferk: Actually, I think the MBF signature prboom carries is still left there, so there are actually 2 signatures in the demo, which I dont think makes much sense:
https://github.com/XaserAcheron/prboom-plus/blob/45b9e0806285d485ff9e1ba7757381021e7f9b24/prboom2/src/g_game.c#L3273-L3279

kraflab: It's layered. The umapinfo data wraps the normal demo, and that following signature determines the compatibility. It could be mbf, but it could also be boom (which has a different signature), etc. This is my understanding anyway. If eternity doesn't include the mbf signature, how will the demo compatibility be detected? It's not assumed that all umapinfo wads are mbf, I assume.

prboom2/src/g_game.c
Altazimuth: OK I've consulted with anotak on the matter. Turns out some demo formats only support map in the form of numbers. This change means we cah have demos without the gameplay you'd have for those formats, but with UMAPINFO support.

XaserAcheron: Ye, basically it's a bit of forward-compatibility, in case UMAPINFO grows to support non-standard map lump names (e.g. "E1A1"). Without this, it'd be impossible to, say, record a -complevel 2 demo in a wad that uses a nonstandard map lump names but is otherwise intended for limit-removing play... and by "impossible" I mean "would fail wildly". :P

Besides, even if we assume the longest case (9 bytes per map lump name) for someone recording a full-megawad playthrough demo of 32 maps, this would only be 288 bytes, which is already less character than this message and is orders of magnitude less data than what you've downloaded by viewing this comment on GitHub. :P

tl;dr, it's necessary to make things future-proof and the extra space it takes is friggin' tiny in 2020.

Altazimuth:
It's been a bit since the last comments. Is there any additional hold-up left on this?

fabiangreffrath:
> It's been a bit since the last comments. Is there any additional hold-up left on this?

It's controversial, to say the least. And the project owner @coelckers has yet to give a sign of acknowledgement.

prboom2/src/g_game.c
fabiangreffrath: Could this question get a bit more elaboration, please? Especially the parts that @Ferk and @kraflab asked are also still unclear to me.

Altazimuth:
I was under the impression al the issues that caused any controversy have been ironed out, and I probably wouldn't wait on Graf. I don't recall him acknowledging this project for at least half a year.

XaserAcheron:
This PR is basically _the_ thing that is stopping this port from being accepted as an "official" successor to prboom-plus mainline, and there are no unaddressed issues here. What's the hold-up?

fabiangreffrath:
> This PR is basically _the_ thing that is stopping this port from being accepted as an "official" successor to prboom-plus mainline, and there are no unaddressed issues here. What's the hold-up?

I somehow feel addressed by this complaint, because apparently I am the only one left with commit rights who is still actively interested in this port.

First of all, this patch is *very* intrusive, given that it all started with the insertion of a single `0xff` byte right before the actual demo signature to indicate that this is a UMAPINFO demo.

Second, this implements what someone else - who is not the author of this fork and thus not the author of the UMAPINFO reference implementation - wants UMAPINFO demo headers to look like. This, without a single sign of acknowledgement by the original author.

Third, there is still an open question regarding the nesting of demo signatures and the necessity of the MBF demo signature. There are questions asked by @Ferk and @kraflab in July and I requested elaboration just yesterday, because - frankly - I fail to understand this part of the code as well.

I understand there is probably not much that can be done about point 1 if we want to keep this level of complexity. Same for point 2 which is nothing that anyone of you could be blamed for. What I'd really like to have, though, is confirmation (a) that UMAPINFO demos can be recorded and played back with *any* complevel and thus embed demos in any other format, (b) that this is forward and backward compatible with The Eternity Engine, and (c) that UMAPINFO demos will be detected and rejected by non-UMAPINFO ports (such as Chocolate).

kraflab:
Would it help if we split this into two pieces?
1) Define / implement the next generation of the demo format, with a flexible and extensible header.
2) Add umapinfo support to that new format.

In particular for (1) it would probably help if there was a (non code) specification of the format to look at to answer the questions raised about compatibility and make sure everyone is on the same page. Maybe that exists on the forum already and could just be linked.

fabiangreffrath:
@kraflab nice idea, though I somehow expected (2) before (1).

prboom2/src/g_game.c
```diff
@@ -3179,13 +3182,74 @@ void G_BeginRecording (void)
 {
   int i;
   byte *demostart, *demo_p;
+  int num_extensions = 0;
   demostart = demo_p = malloc(1000);
   longtics = 0;
 
-  if (umapinfo_loaded)
+  // ano - jun2019 - add the extension format if needed
+  if (gamemapinfo)
```
fabiangreffrath, line 3190: Why is this changed from `umapinfo_loaded` to `gamemapinfo` here? AFAIUI, `gamemapinfo` is only set if there is UMAPINFO information available _for the current map_, whereas `umapinfo_loaded` is set a soon as a `UMAPINFO` lump is found and loaded. But what if we start a demo from a map without UMAPINFO changes and then migrate to UMAPINFO affected maps during the course of recording?

prboom2/src/g_game.c
```diff
@@ -3512,21 +3578,175 @@ const byte* G_ReadDemoHeaderEx(const byte *demo_p, size_t size, unsigned int par
...
+    // so that's a total of 1+6+1+2 + (n * m) bytes + ?? for extensions
+    // or 10 + some ?? bytes + some more ??
+
+    // then finally the "real" demover byte is present here
+
+    if (CheckForOverrun(header_p, demo_p, size, 10, failonerror))
+      return NULL;
+
+    // we check for the PR+UM signature as mentioned.
+    // MBF and Eternity Engine also use 255 demover, with other signatures.
```
fabiangreffrath, line 3614: I think MBF uses 203 as its demo version number. Please double-check and correct here if this is true or tell me if I'm wrong. ;)

prboom2/src/g_game.c
```diff
@@ -3512,21 +3578,175 @@ const byte* G_ReadDemoHeaderEx(const byte *demo_p, size_t size, unsigned int par
...
+          {
+            if (cur == 0)
+            {
+              break;
+            }
+
+            I_Error("G_ReadDemoHeader: Unable to determine map for UMAPINFO demo.");
+          }
+
+          episode = (episode * 10) + (cur - '0');
```
fabiangreffrath, line 3731: This should be `map` here instead of `episode`, right?

fabiangreffrath (changes requested):
I gave this patch another review.

In general, the part in `G_BeginRecording()` looks good to me, but raises a question as to when the special UMAPINFO treatment gets triggered (if UMAPINFO lump found and loaded vs. if demo start map has UMAPINFO changes applied).

Then there is a either misconception about MBF using `0xff` as its demo version number, or I am missing something.

Third, there is a copy/paste bug in the `G_ReadDemoHeaderEx()` part that needs to be addressed - and that leaves me wondering how much this code has actually been tested in real life scenarios.

Please address these issues before we can proceed, thanks!

prboom2/src/g_game.c
XaserAcheron: Good catch -- it indeed oughta be `umapinfo_loaded`. I whipped up a test wad which contained UMAPINFO for maps 2, 3, and 4, recorded a demo starting from MAP01, and it skipped writing the extension header entirely. Got a fix incoming.

prboom2/src/g_game.c
XaserAcheron: Fixed. :P

prboom2/src/g_game.c
XaserAcheron: Yup. Fixed.

XaserAcheron:
Finally had some time to sit down and play around with this some more -- tl;dr: made all the requested changes.

I put together a quick test wad while I was at it to repro & fix the troubles:
[umapinfo-test.zip](https://github.com/coelckers/prboom-plus/files/5741467/umapinfo-test.zip)

The wad contains a couple of maps, + a UMAPINFO file that defines maps 02, 03, 33, and 04 (but not MAP01), and sets up a branching path to MAP03/MAP33 via secret exit on MAP02.

I recorded  `umapinfo-test-demo-bad.lmp` before the requested changes, and sure enough, it lacks the extension header. I was able to play it back on mainline PRBoom-Plus, where it desynced after hitting the secret exit switch in MAP02. Meanwhile, `umapinfo-test-demo-good.lmp` was recorded after the fix, and it's got the new header in place (non-forked PRB+ correctly aborts with an error). Both play back as expected on this branch, but it's the abort-on-old-versions behavior we're really after here.

fabiangreffrath:
@XaserAcheron Thank you very much for taking care of the requested changes!

One last thing that still annoys me is the completely redundant parsing of the ExMy/MAPxy level name. I know, the reason is that it should be possible in the future to have non-standard map names defined by UMAPINFO. However, the current implementation does not support this and even UMAPINFO's own `G_ValidateMapName()` [1] does not support it either. So, it is very unlikely to happen anytime soon and I'd rather prefer to have this entirely removed and delayed for a v2 format specification.

Anyway, it seems to work as intended now and I am not going to argue about this PR anymore. 

Merging, merry (late) Christmas everybody! :christmas_tree: 

[1] https://github.com/XaserAcheron/prboom-plus/blob/7882c46da75554f9ffa996acded647904d867167/prboom2/src/g_game.c#L2692

PR #114, opened 2020-08-12
Add install rules for prboom-plus documentation

vilhelmgray:


fabiangreffrath:
TTBOMN, manual pages should not get installed directly under `/usr/share/man` but into section subdirectories instead, e.g. `/usr/share/man/man5/prboom-plus.cfg.5` and `/usr/share/man/man6/prboom-plus.6`.

https://refspecs.linuxfoundation.org/FHS_3.0/fhs-3.0.html#usrsharemanManualPages

vilhelmgray:
Thanks for the link, I learned something new today about manpages. I've updated the commit to install the manpages under their appropriate subdirectories. :-)

(By the way, I feel a bit out of the loop to have to ask, but what does TTBOMN mean?)

prboom2/CMakeLists.txt
```diff
@@ -183,5 +183,20 @@ set(PRBOOM_OUTPUT_PATH ${CMAKE_BINARY_DIR})
...
+	set( DOCDIR "${CMAKE_INSTALL_PREFIX}/share/doc/${PROJECT_NAME}" CACHE STRING "Directory to install documentation into" )
+endif()
+install(DIRECTORY doc/ DESTINATION ${DOCDIR} COMPONENT "Documentation")
+
+if( WIN32 )
+	set( MANDIR man CACHE STRING "Directory to install manpages into" )
+else()
+	set( MANDIR "${CMAKE_INSTALL_PREFIX}/share/man" CACHE STRING "Directory to install manpages into" )
+endif()
+install(DIRECTORY man/man5/ DESTINATION "${MANDIR}/man5" COMPONENT "Man5 manpages")
```
fabiangreffrath, line 198: Couldn't we just `install(FILES doc/*.5 ...` or similar here? Then we souldn't have to move the files around at all.

fabiangreffrath:
> (By the way, I feel a bit out of the loop to have to ask, but what does TTBOMN mean?)

to the best of my knowledge 😁

prboom2/CMakeLists.txt
vilhelmgray: I don't believe wildcards are allowed when using `install(FILES ...`; I could list each file out explicitly. However, this does present the problem that the manpages would be installed in two locations because they will also be picked up by the earlier `install(DIRECTORY doc/ ...` line.

vilhelmgray: Do you think I should list out all the documentation files explicitly -- I think this is probably the style CMake is expecting in fact, and it'll prevent the double installation.

vilhelmgray:
I discovered the CMake `install` command supports a `TYPE` configuration that allows us to specify whether the files are documentation or manpages. So I've updated the commit to use that rather than rolling our own `DOCDIR`/`MANDIR`.

fabiangreffrath:
Any chance for a similarly elegant solution that doesn't require to move the manpage files around?

vilhelmgray:
I'm not especially experienced in CMake, but from what I gathered from the documentation, if we do not move the manpage files we will need to solve two problems:

* Prevent the unintentional installation of documentation files into manpage system directories (and vice-versa).
* Install manpage files into their respective `man5` and `man6` system directories.

The first problem can be solved by using `FILES` option to specify each file individually. The second problem can be solved by rolling our own custom MANDIR and specifying the manpage files destinations explicitly. This would result in the following alternative CMake code:

```CMake
install(FILES doc/boom.txt TYPE DOC COMPONENT "Documentation")
install(FILES doc/DeePBSPV4specs.txt TYPE DOC COMPONENT "Documentation")
install(FILES doc/MBFFAQ.txt TYPE DOC COMPONENT "Documentation")
install(FILES doc/MBF.txt TYPE DOC COMPONENT "Documentation")
install(FILES doc/prboom-plus-history.html TYPE DOC COMPONENT "Documentation")
install(FILES doc/prboom-plus-history.txt TYPE DOC COMPONENT "Documentation")
install(FILES doc/prboom-plus-usage.txt TYPE DOC COMPONENT "Documentation")
install(FILES doc/README.command-line TYPE DOC COMPONENT "Documentation")
install(FILES doc/README.compat TYPE DOC COMPONENT "Documentation")
install(FILES doc/README.demos TYPE DOC COMPONENT "Documentation")
install(FILES doc/umapinfo.txt TYPE DOC COMPONENT "Documentation")

if( WIN32 )
        set( INSTALL_MAN_PATH man CACHE STRING "Directory to install manpages into" )
else()
        set( INSTALL_MAN_PATH ${CMAKE_INSTALL_PREFIX}/share/man CACHE STRING "Directory to install manpages into" )
endif()
install(FILES man/prboom-plus.cfg.5 DESTINATION "${INSTALL_MAN_PATH}/man5" COMPONENT "Manpages")
install(FILES man/prboom-plus.6 DESTINATION "${INSTALL_MAN_PATH}/man6" COMPONENT "Manpages")
install(FILES man/prboom-plus-game-server.6 DESTINATION "${INSTALL_MAN_PATH}/man6" COMPONENT "Manpages")
```

This alternative solution seems a lot more complex than the current solution, so moving the manpages might be the best way path forward in this case.

fabiangreffrath:
You can add files to variables pretty easily, see e.g. the `prboom2/data/CMakeLists.txt` file. Then we could have three sets of files, i.e. the docs, the man5 and the man6 files, and install them with three `install(FILES ...)` rules.

vilhelmgray:
> You can add files to variables pretty easily, see e.g. the `prboom2/data/CMakeLists.txt` file. Then we could have three sets of files, i.e. the docs, the man5 and the man6 files, and install them with three `install(FILES ...)` rules.

Hmm, maybe that might work. I'll play around with it later and see if I can come up with something succinct.

fabiangreffrath:
Maybe you should rather move these entire rules into a separate `doc/CMakeList.txt` file and include this from the top-level file.

vilhelmgray:
Here's what I've come up with (`doc/CMakeLists.txt`):

```CMake
include(GNUInstallDirs)

set(DOC_FILES
    boom.txt
    DeePBSPV4specs.txt
    MBFFAQ.txt
    MBF.txt
    prboom-plus-history.html
    prboom-plus-history.txt
    prboom-plus-usage.txt
    README.command-line
    README.compat
    README.demos
    umapinfo.txt
)
set(MAN5_FILES
    prboom-plus.cfg.5
)
set(MAN6_FILES
    prboom-plus.6
    prboom-plus-game-server.6
)

install(FILES ${DOC_FILES} TYPE DOC COMPONENT "Documentation")
install(FILES ${MAN5_FILES} DESTINATION "${CMAKE_INSTALL_MANDIR}/man5" COMPONENT "Manpages")
install(FILES ${MAN6_FILES} DESTINATION "${CMAKE_INSTALL_MANDIR}/man6" COMPONENT "Manpages")
```

I'm including [GNUInstallDirs](https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html) so that we don't need to roll our own `MANDIR`. Let me know what you think; if you like this approach better, I'll force-push to update the commit.

fabiangreffrath:
That's what I had in mind! :+1:

fabiangreffrath:
Thanks!

Issue #118, opened 2020-09-05
[Bug] Game exits after text screen


Porcozo:
After exiting MAP06 in interception 2, the game exits with this message:
![Errormessage](https://user-images.githubusercontent.com/47692609/92310487-5d1b9900-ef85-11ea-9e0e-1692d4cd366f.png)

Here's the corresponding UMAPINFO lines:
map MAP06
{
	levelname = "What Could Go Wrong?"
	levelpic = "CWILV05"
	next = "MAP07"
	nextsecret = "MAP31"
	skytexture = "SKY1"
	music = "D_THE_DA"
	exitpic = "INTERPIC"
	enterpic = "INTERPIC"
	partime = 0
	endgame = false
	intertext = "You refuse to let the secrets of the under",
		"world corrupt your mind.",
		"",
		"At least that's what you've been telling",
		"yourself since you got here. The demons",
		"have a beautiful way of whispering into",
		"your psyche's deepest reaches...",
		"",
		"In short, it's a miracle you haven't",
		"been transformed into one of them. And",
		"yet, it remains harder than ever to",
		"recall how long it's been since you left",
		"Earth..."
	interbackdrop = "GSTONEF"
	intermusic = "D_READ_M"
	intertextsecret = clear
}

I'm using seed's build from july 1st.

JadingTsunami:
Looks like a real bug to me. The line below I believe has the boolean check inverted.

```
prboom2/src/umapinfo.cpp:362
	else if (!stricmp(pname, "endgame"))
	{
		scanner.MustGetToken(TK_BoolConst);
/* this if check should be for !scanner.boolean */
		if (scanner.boolean) strcpy(mape->endpic, "!");
		else strcpy(mape->endpic, "-");
	}
```

JadingTsunami:
This appears to be a bug to me. Putting "endgame = false" but no endpic will cause this crash, whereas I believe the correct behavior would be to proceed to the next level.

I believe this is caused by the fact that the finale handler will check for the presence of any nonzero character in the first position of the endpic string as evidence that the endgame should be triggered, but "endgame = false" will copy a "-" into endpic, thereby triggering those conditions to evaluate true.

The fix is either to nullify the first character of endpic when the "endgame = false" token is parsed from the UMAPINFO lump, or modify the endpic checks to look for "!= '-'" instead of a null check on the zeroth character of endpic.

[...]

PR #119, opened 2020-09-06
Update endpic checks to also check for '-' (blank) endpic.

JadingTsunami:
A proposed fix for #118.

The UMAPINFO scanner will insert a '-' character to indicate a blank endpic. However, the in-game checks for endpic only check for a null endpic, which would happen if UMAPINFO is loaded but no endpic or endgame entries are found and the map is not normally an endgame map.

This PR proposes a fix to this by modifying gameplay checks to look for a '-' character endpic. This would occur if endgame was set to false, or a traditional cast call or "bunny" ending was falsified.

### DIFF
--- a/prboom2/src/f_finale2.c
+++ b/prboom2/src/f_finale2.c
@@ -124,7 +124,7 @@ void FMI_Ticker(void)
 			(midstage && acceleratestage)) {
 
 		next_level:
-			if (gamemapinfo->endpic[0])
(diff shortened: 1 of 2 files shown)


fabiangreffrath:
Looks reasonable. The `gamemapinfo->endpic` string must not be empty, but it must not be the dash string "-" either.

fabiangreffrath (approved):

PR #123, opened 2020-10-02
Fix segfault when handling -viewangle parameter when it is the last parameter

onContentStop:
I found a segfault in the parameter handling code. A small code change should do it.

By the way, this is my first contribution; please be nice :) and let me know if I'm missing anything.


fabiangreffrath (approved):

fabiangreffrath:
Thank you!

PR #124, opened 2020-10-03
Do not reset next levels if endscreen is disabled in UMAPINFO

RRFGamer:
If the `endpic` property is present in an UMAPINFO lump, the `next` and `nextsecret` properties are *zero'ed*, even if the end screen is disabled by `endpic = "-"`. This may mess up intended level progression. Building with this fix solves the problem.

I didn't include an exception for "!" since i don't fully understand the behaviour of `endgame`.
In this case, `next` and `nextsecret` are still *zero'ed*, but the game only ends when that's the default (like in MAP30).

### DIFF
--- a/prboom2/src/umapinfo.cpp
+++ b/prboom2/src/umapinfo.cpp
@@ -520,7 +520,7 @@ int ParseUMapInfo(const unsigned char *buffer, size_t length, umapinfo_errorfunc
 		ParseMapEntry(scanner, &parsed);
 
 		// Set default level progression here to simplify the checks elsewhere. Doing this lets us skip all normal code for this if nothing has been defined.
-		if (parsed.endpic[0])
+		if (parsed.endpic[0] && parsed.endpic[0] != '-')
 		{
 			parsed.nextmap[0] = parsed.nextsecret[0] = 0;
 			if (parsed.endpic[0] == '!') parsed.endpic[0] = 0;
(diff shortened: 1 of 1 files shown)


JadingTsunami:
As a general comment, I recommend doing a string comparison instead of a single character, because '-' is a legal character in lumps. It is possible (albeit unlikely) that an endpic lump could begin with '-' in which case this check would fail incorrectly.

For this specific fix, I think this is not necessary if the UMAPINFO specification were updated (or maybe clarified) to note that only end game option per map is allowed. E.g., "endgame = false" would be the expected way to override default endgame behavior for a given map. But others may have more thoughts on this as well.

fabiangreffrath (approved):

JadingTsunami:
@fabiangreffrath I see the changes are approved, but I'd still recommend at least making this code consistent with the other checks for a blank endpic and making it a string comparison instead of a 0th character check only.

I would either move all the checks to be 0th character only and create a new rule that prohibits pic lumps from starting with '-' or make this check match the others and perform a proper string comparison for "-" as a full string.

fabiangreffrath:
I approved the PR and then requested the code changes that you suggested for code consistency.

fabiangreffrath (changes requested):
Please change this to a full string comparison as suggested by @JadingTsunami and I'll merge it.

fabiangreffrath:
Thank you!

RRFGamer:
Thanks for the patience. Took me a while to figure out git-scm.

Issue #125, opened 2020-10-07
CMake: Install files given no destination


zerojay:
I'm getting the following error when attempting to compile on a Pi 4:

mkdir build
cd build
cmake -DCMAKE_BUILD_TYPE=Release ..

....
-- Looking for SDL_JoystickGetAxis - found
CMake Error at doc/CMakeLists.txt:22 (install):
  install FILES given no DESTINATION!

fabiangreffrath:
@vilhelmgray Could you please take a look?

vilhelmgray:
Looks like it's getting confused on this line:
https://github.com/coelckers/prboom-plus/blob/f404802515cf26ce933e54873d71ee9aa29540ff/prboom2/doc/CMakeLists.txt#L22

I suspect CMake is getting stuck on the [`TYPE` parameter of `install(FILES ...)`](https://cmake.org/cmake/help/latest/command/install.html#installing-files). This could happen if you're running an old version of CMake, or your CMake installation has some sort of bug.

Let's try to track down your CMake setup so that we can reproduce this issue and troubleshoot it. What version of CMake are you running (try running `cmake --version`)?

fabiangreffrath:
No matter what version @zerojay uses, could we just revert the offending part of the Cmake rule to make it work with any older version, please?

vilhelmgray:
Sure I can put a conditonal to check for the CMake version, but we still need to determine if there is a bug occurring here in the CMake system installation itself.

zerojay:
cmake --version

cmake version 3.13.4

This is what Raspbian/Raspberry Pi OS (Buster) is currently shipping with.

vilhelmgray:
@zerojay Thanks, it looks like this is just a case of an old CMake that doesn't support the newer installation option. I'll write a quick patch to add basic support for the older CMake versions.

vilhelmgray:
@zerojay Please test out https://github.com/coelckers/prboom-plus/pull/126 and let me know if this resolves the issue for you. This should use the older `DESTINATION` parameter for CMake versions less than the [3.14.0 release where support for the `TYPE` parameter was added](https://cmake.org/cmake/help/v3.14/release/3.14.html#commands).

zerojay:
Looks good, cmake completes and make compiles. Thank you for the incredibly quick turnaround, everyone.

fabiangreffrath:
[...]

PR #126, opened 2020-10-07
Provide documentation installation support for older CMake

vilhelmgray:
Support for the `TYPE` parameter of the CMake `install(FILES ...)`
command was added in the version 3.14.0 release of CMake. Older versions
should use the `DESTINATION` parameter as an alternative.

Bug: https://github.com/coelckers/prboom-plus/issues/125


fabiangreffrath:
Thanks guys!

PR #127, opened 2020-10-13
200 Sounds for DEHEXTRA

Shadow-Hog:
To go along with the extra things, states and sprites.

Numbers chosen after discussion with Altazimuth over Eternity Engine already using the next slots for _Heretic_ and _Strife_ sounds.  (See: [https://www.doomworld.com/forum/topic/115638-dehextra-discussion-split-from-pr-umapinfo-thread/](url))

For testing purposes, [three sample files are also attached](https://github.com/coelckers/prboom-plus/files/5368976/DSFRExxSounds.zip):
- DSFRExxxSounds.wad: A computer counts down every two seconds from 200, and you die upon it hitting zero.
- DSFRExxxSounds_hyperspeed.wad: A computer counts down every two tics from 200, and you die upon it hitting zero.
- DSFRExxxSounds_hyperspeed_nosounds.wad: A computer tries to count down every two tics from 200, but the lumps aren't in the WAD file (just to make sure the program handles missing lumps correctly) so it remains completely silent. Regardless, upon it hitting zero, you die.

Also tested what happens if you try to play a sound in the newly-introduced gap in the definition table - no sound plays, but also no errors are thrown/no crashes occur, which is the desired behavior.


prboom2/src/sounds.c
```diff
@@ -247,4 +247,210 @@ sfxinfo_t S_sfx[] = {
   //e6y
   { "secret", false, 60, 0, -1, -1, 0 },
   { "gibdth", false, 60, 0, -1, -1, 0 },
+  // Everything from here up to 400 is reserved for future use.
```
fabiangreffrath, line 250: Or 500?

prboom2/src/sounds.c
```diff
@@ -247,4 +247,210 @@ sfxinfo_t S_sfx[] = {
...
+
+  // Free slots for DEHEXTRA. Priorities should be overridden by user.
+  // The gap to accomodate Eternity means I must use desginated initializers,
+  // or otherwise supply an exact number of dummy entries to pad it out.
+  // Not sure which would be uglier to maintain. -SH
+  [500] = { "fre000", false, 127, 0, -1, -1, 0 },
```
fabiangreffrath, line 256: Hm, is the "freXXX" name scheme agreed upon with other ports?

fabiangreffrath (changes requested):
LGTM! Could you please rebase your changes on the current master branch?

fabiangreffrath:
Could you probably reply to the two questions I raised above, please?

prboom2/src/sounds.c
Shadow-Hog: Yeah, that's a silly oversight - it _was_ 400 until shortly before the commit, when Altazimuth requested another 100 reserved slots for EE. Either way, fixed.

prboom2/src/sounds.c
Shadow-Hog: I posted the naming scheme in the DEHEXTRA thread on Doomworld hoping to receive comment, but only really heard back from Altazimuth of Eternity Engine, who I believe has already implemented the sounds with these lump names. Who else should I pester?

Shadow-Hog:
Sorry, got caught up in figuring out if I _could_ rebase after the merge that I didn't take care of the other two issues first.

As for that, it's _probably_ possible, but I don't know how easy it is to do off-hand. I know I tried rebasing to begin with, but every rebase would introduce file conflicts in files I didn't even touch, dragging in DECOLITE code from God-knows-where. (Hell, even the _merge_ did that...) Any advice on this would be helpful, my local repo somehow got desynced in 2017 and it's been... _fun_ trying to get it back up to speed.

prboom2/src/sounds.c
fabiangreffrath: I see. Could you probably add a link to the relevant part of the thread in the comment? Or a citation of Altazimuth's reply? Or a link to the commit which adds this code to the Eternity Engine?  Because without this knowledge, the chosen naming scheme leaves me a bit puzzled.

fabiangreffrath:
I guess we can get over the rebasing chaos by doing a squash merge.

prboom2/src/sounds.c
fabiangreffrath: So?

Shadow-Hog (commented):
I'm very confused. I left a comment answering your request for links the same day you asked it, but Github thus said it was "Pending", and only today do I learn that it hid it from everyone but me, and won't explain why or how to get it to actually show up. Maybe clicking "Resolve Conversation" will fix that? I thought that would prematurely shut the conversation based on the verbiage, but maybe it doesn't.  (The only other option is to just uselessly edit the comment, with no option to move it out of Pending free there.)

EDIT: No, that definitely does what I thought it did. Unresolved it, but hey, the pending comment is no longer pending now.

prboom2/src/sounds.c
fabiangreffrath: Hm, forum links are pretty long these days. I guess it would be best to add something like "This implements prior art added to The Eternity Engine in commit https://github.com/team-eternity/eternity/commit/b8fb8f71." or similar to the comment (I'm not a native speaker, so please bear with me). Then I'd be fine with this PR.

Shadow-Hog: Comment adjusted to include the URL of that commit.

fabiangreffrath:
Merged, thank you for your patience!

PR #139, opened 2020-11-26
r4552 | entryway | 2020-09-02 05:33:49 +0200 (Wed, 02 Sep 2020) | 1 line

fabiangreffrath:
"No Vertical Mouse" setting and keybind

### DIFF
--- a/prboom2/src/e6y.c
+++ b/prboom2/src/e6y.c
@@ -133,6 +133,7 @@ int hudadd_crosshair_lock_target;
 int movement_strafe50;
 int movement_shorttics;
 int movement_mouselook;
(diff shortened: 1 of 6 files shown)


kraflab:
So, old prboom+ is still getting updates sometimes? What happens if it's updated to 2.5.1.7 at some point? 😅

fabiangreffrath:
> So, old prboom+ is still getting updates sometimes? What happens if it's updated to 2.5.1.7 at some point? 😅

Yes, though only very occasionally - only 10 commits this year and half of them were backports of our changes here. If it ever reaches 2.5.1.7, I hope we are at 3.0 already. 😉

PR #140, opened 2020-11-27
weapon change sequence complete when A_WeaponReady() is called

### LINKED ISSUE #131: Cannot draw weapons, stuck on angry face
I don't have the exact steps to reproduce this bug but will provide them once I figure them out.

**Expected:**
Can always switch weapons.

**Actual:**
Sometimes doomguy's face "gets stuck" on the angry face (as in defaults to it) and can no longer draw any weapon.
By that I mean that I'm pressing 1-7 and nothing happens. No weapon is displayed either. Can't shoot.

**Extra:**
Saving and loading the game on the same session doesn't seem to set it back to normal behaviour.
Toggling IDDQD or IDCLIP when the bug kicks in doesn't seem to do anything useful.
Dying seems to reset it back to normal though.

**Edit:**
Save file (change extension from 'docx' to 'dsg')
[prboom-plus-savegame0.docx](https://github.com/coelckers/prboom-plus/files/5516800/prboom-plus-savegame0.docx)
Restarting the program and loading the save doesn't restore normal behaviour.

Died a good few times due to this.

The problem apparently started since I compiled  534f9552 and haven't experienced it back when I used 47ebaf83 or so (or maybe just didn't take much note of it back then).

fabiangreffrath:
Fixes #131.

### DIFF
--- a/prboom2/src/p_pspr.c
+++ b/prboom2/src/p_pspr.c
@@ -329,6 +329,9 @@ void A_WeaponReady(player_t *player, pspdef_t *psp)
 {
   CHECK_WEAPON_CODEPOINTER("A_WeaponReady", player);
 
+// weapon change sequence considered complete
+  done_autoswitch = false;
+
   // get out of attack state
   if (player->mo->state == &states[S_PLAY_ATK1]
       || player->mo->state == &states[S_PLAY_ATK2] )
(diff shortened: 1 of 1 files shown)


kraflab:
Thanks to everybody that investigated this :+1:

PR #144, opened 2020-11-30
- moved the secret notification to the "Miscellaneous" submenu inside "General"

sinisterseed:
Mainly to improve visibility and awareness, I've seen people not knowing this even exits some time ago, including yours truly.

Rightfully so, its original location inside "Setup" (really, Setup out of all things?) is not the best one could come up with, admittedly. I think it fits much better here.

I've also renamed it to better fit in with the rest of the settings. I initially got it half-right though, hence two commits - it was still trying to display like the settings from its original location, because I missed the obvious, heh.

Now it's good, and of course it's tested. It works as it should.

### DIFF
--- a/prboom2/src/m_menu.c
+++ b/prboom2/src/m_menu.c
@@ -3198,11 +3198,11 @@ setup_menu_t gen_settings2[] = { // General Settings screen2
   {"Game speed, percentage of normal", S_NUM|S_PRGWARN, m_null, G_X, G_Y+13*8, {"realtic_clock_rate"}},
   {"Default skill level",              S_CHOICE,        m_null, G_X, G_Y+14*8, {"default_skill"}, 0, 0, NULL, gen_skillstrings},
   {"Default compatibility level",      S_CHOICE,        m_null, G_X, G_Y+15*8, {"default_compatibility_level"}, 0, 0, NULL, &gen_compstrings[1]},
-  {"Show Secret Notifications",        S_YESNO,         m_null, ADVHUD_X, SB_Y+ 2*8, {"hudadd_secretarea"}},
-  {"Show ENDOOM screen",               S_YESNO,         m_null, G_X, G_Y+16*8, {"showendoom"}},
-  {"Fullscreen menu background",       S_YESNO, m_null, G_X, G_Y + 17*8, {"menu_background"}},
+  {"Show Secret Notifications",        S_YESNO,         m_null, G_X, G_Y+16*8, {"hudadd_secretarea"}},
+  {"Show ENDOOM screen",               S_YESNO,         m_null, G_X, G_Y+17*8, {"showendoom"}},
+  {"Fullscreen menu background",       S_YESNO, m_null, G_X, G_Y + 18*8, {"menu_background"}},
 #ifdef USE_WINDOWS_LAUNCHER
-  {"Use In-Game Launcher",             S_CHOICE,        m_null, G_X, G_Y+ 18*8, {"launcher_enable"}, 0, 0, NULL, launcher_enable_states},
+  {"Use In-Game Launcher",             S_CHOICE,        m_null, G_X, G_Y+ 19*8, {"launcher_enable"}, 0, 0, NULL, launcher_enable_states},
 #endif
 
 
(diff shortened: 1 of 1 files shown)


kraflab:
This is a hud topic though, doesn't it make more sense to be with the other options for this area, where it was? Albeit not advanced hud, but that page also has crosshair, another hud element.

prboom2/src/m_menu.c
```diff
@@ -2720,7 +2720,6 @@ setup_menu_t stat_settings1[] =  // Status Bar and HUD Settings screen
 setup_menu_t stat_settings2[] =
 {
   {"ADVANCED HUD SETTINGS"       ,S_SKIP|S_TITLE,m_null,ADVHUD_X,SB_Y+1*8},
-  {"REPORT REVEALED SECRETS"     ,S_YESNO     ,m_null,ADVHUD_X,SB_Y+ 2*8, {"hudadd_secretarea"}},
```
kraflab, line 2723: You'll need to adjust the indices of the options below this one to shift them up to fill the gap if we remove from here.

sinisterseed: Yeah... I realized this just now while I was reviewing the changes. _I had one job._

I accidentally left an empty white space there now, damn. Not sure what to make of it though, it seems some aren't fond of this change, so I'd rather just close this and call it a day.

sinisterseed:
Actually, I may have another idea.

Should we just port the advanced HUD settings to general instead? On a new page (or more, haven't checked how many of them are there)?

Them being in General feels a lot better to me than Setup. Or could be just rename Setup to something else that actually makes some sense?

prboom2/src/m_menu.c
```diff
@@ -3199,10 +3198,11 @@ setup_menu_t gen_settings2[] = { // General Settings screen2
   {"Game speed, percentage of normal", S_NUM|S_PRGWARN, m_null, G_X, G_Y+13*8, {"realtic_clock_rate"}},
   {"Default skill level",              S_CHOICE,        m_null, G_X, G_Y+14*8, {"default_skill"}, 0, 0, NULL, gen_skillstrings},
   {"Default compatibility level",      S_CHOICE,        m_null, G_X, G_Y+15*8, {"default_compatibility_level"}, 0, 0, NULL, &gen_compstrings[1]},
-  {"Show ENDOOM screen",               S_YESNO,         m_null, G_X, G_Y+16*8, {"showendoom"}},
-  {"Fullscreen menu background",       S_YESNO, m_null, G_X, G_Y + 17*8, {"menu_background"}},
+  {"Show Secret Notifications",        S_YESNO,         m_null, G_X, G_Y+16*8, {"hudadd_secretarea"}},
```
fabiangreffrath, line 3201: Maybe that's at least one thing we can get out of this PR after all. Does "Show Secret Notifications" make more sense to a native speaker than "REPORT REVEALED SECRETS"? Is it immediately clear what is meant? I am open to rename the menu item once again if there is consensus about an improved wording.

fabiangreffrath: I mean, the notifications are not secret, they are about secrets. But maybe that's just me?

sinisterseed: No, I agree that it sounds much better to me as well.

sinisterseed: BTW what about my other proposal, of moving the automap settings to a new page under General, or renaming Setup to something more meaningful?

fabiangreffrath: If we agree to overhaul the menu, we should do it as a whole and not move a single section. Once you _know_ where to find a setting, it doesn't matter that much anymore how hard it was to find it. What we should probably avoid is to move some settings to a new logical place but leave others where they were. I am open to renaming some of the sections, though I think that "Setup" is quite fitting for settings things like the display resolution.

sinisterseed: Yes, Setup in itself is not a poor name, the issue is what it currently houses. Stuff like HUD and automap settings is the more glaring issue in my book, instead of housing resolution, FOV, compatibility, the ENDOOM screen, and the list goes on.

Honestly I'd be just fine with an overhaul, but I'm not even sure where to begin with something so grand in scale. Any suggestions?

Issue #145, opened 2020-12-07
Signal 11 with "Sunder 2308.wad"


kraflab:
https://www.doomworld.com/forum/topic/46002-sunder-map19-arrives/

Reported by Shepardus on discord.

Loading this wad in prboom+ or glboom+ on windows causes a signal 11. It _does not happen on linux_.

original 2.5.1.7 (June 16 2019): loads fine

I went as far back as commit 8a390a16853aef9348d714eb3b0b4e15f39ff217 (May 12 2020) and the error is there as well. I can't compile the older commits currently because that's when gcc 10.1 compilation was fixed :^)

So, somewhere in between June 16 2019 and May 12 2020 there was a change that broke something on windows.

fabiangreffrath:
Hm, cannot seem to reproduce, both with `-warp 19` and with IDCLEV19.

Does music start to play? How about `-nomusic`?

Is anything useful printed to the terminal?

kraflab:
It happens before the game window even appears - that is, just from loading the wad, not even entering a map. It happens with `-nomusic` as well. I thought it could be an issue with how I compile it, but it happens with one of the builds another user posted in the dw thread, so I guess that's not it.

fabiangreffrath:
Hm, still cannot reproduce here... :/

fabiangreffrath:
Since you are using GCC, do you have a chance to run the executable in GDB and report where it crashes?

`gdb --args ./prboom-plus -file "Sunder 2308.wad"`, and then inside GDB type `r` to run the program and once it crashes type `bt` to get a backtrace.

kraflab:
```
#0  0x00007fff064313ad in ntdll!RtlEnterCriticalSection () from C:\WINDOWS\SYSTEM32\ntdll.dll
#1  0x00007fff0484ccfe in msvcrt!fgets () from C:\WINDOWS\System32\msvcrt.dll
#2  0x0000000000408746 in dehfgets ()
#3  0x00000000004095e2 in ProcessDehFile ()
#4  0x00000000004118de in D_DoomMainSetup ()
#5  0x00000000004120a6 in D_DoomMain ()
#6  0x00000000004f564f in SDL_main ()
#7  0x000000000050bca9 in main_getcmdline ()
#8  0x000000000050bd4b in WinMain ()
```

Looks like that is this line in `d_deh.c`:
```
  // loop until end of file

  while (dehfgets(inbuffer,sizeof(inbuffer),filein))
```

[...]

PR #146, opened 2020-12-07
Fix empty dehacked lumps

kraflab:
Fixes https://github.com/coelckers/prboom-plus/issues/145


fabiangreffrath:
Just changed it in Woof to perform only the second check and it works just as well. Sorry for the bad timing.

fabiangreffrath:
Top, thanks!

PR #151, opened 2020-12-17
Update INSTALL

crabcrabcam:
I have changed the "Basic Installation" to the top as the SVN isn't the main place I believe people get the code from, and changed the Basic Install guide to have the correct instructions (with ./configure missing and instead being cmake .) and added some basic instructions for manual installation (copying the files instead of using make install) as well as a link to FreeDoom as a potential alternative if the user doesn't have iD's Doom. 

The required libraries are correct on Ubuntu and Debian systems (tested on Ubuntu 20.04) and with minor changes to packages downloaded, these instructions work on Void Linux too.

### DIFF
--- a/prboom2/INSTALL
+++ b/prboom2/INSTALL
@@ -29,7 +29,7 @@ Basic Installation (ZIP, tarball, Git)
      Run `cmake .. -DCMAKE_BUILD_TYPE=Release` from this folder, or 
      `cmake .. -DCMAKE_BUILD_TYPE=Debug` for a debug build. (This gives
      slightly lower performance, but more details in the event of 
-     issues)
+     issues).
      
      This should end with the message "Build files have been written to: [folderpath]".
 
@@ -59,7 +59,7 @@ Basic Installation (ZIP, tarball, Git)
      can access the IWAD file.
      
      You may also aquire FreeDoom from https://freedoom.github.io/ which is a 
-     free replacement to the iD Doom iWADs
+     free replacement to the iD Doom iWADs.
 
   6. You can remove the program binaries and object files from the
      source directory by typing `make clean`.  To also remove the
@@ -73,31 +73,6 @@ Basic Installation (ZIP, tarball, Git)
   7. PrBoom is now ready for use. If /usr/local/games is in your path,
      you can just run prboom; otherwise give the full path.
 
-     
-     
-Installation From SVN
-=====================
-
-If you intend to compile PrBoom from a source tar file, such as
-prboom-plus-2.5.1.5.tar.gz, as opposed to SVN, skip to the "Basic Installation"
-section.
-
-  1. Extract the PrBoom distribution from SVN:
-        svn co https://svn.prboom.org/repos/branches/prboom-plus-24/prboom2
-
-  2. 'cd' to the directory containing the PrBoom distribution (the directory
-     this file is in):
-        cd prboom2
-
-  3. Run the 'bootstrap' script in order to create the 'configure' script.
-        ./bootstrap
-     If you get this error:
-        aclocal: configure.ac: 153: macro `AM_PATH_SDL' not found in library
-     Then you have to locate sdl.m4 and copy it to ./autotools/sdl.m4 in your
-     prboom directory.
-
-  4. If the 'configure' script was created by step 3 then proceed to the
-     "Basic Installation" section.
 
      
 Building an RPM
--- a/prboom2/INSTALL
+++ b/prboom2/INSTALL
@@ -1,3 +1,77 @@
+Basic Installation (ZIP, tarball, Git)
+==================
+
+  1. To compile PrBoom-Plus, you need at least SDL2. To enable various
+     extensions and optional features, the following are required:
+     - SDL2_net (for network play)
+     - SDL2_image (for high-res textures in OpenGL, and PNG screenshots)
+     - PCRE (for demo filename pattern matching)
+     - for music support, some or all of
+       * SDL2_mixer (midi, timidity)
+       * Fluidsynth (midi)
+       * Portmidi (midi)
+       * MAD (mp3)
+       * DUMB (various tracker-style formats)
+       * vorbisfile (ogg)
+     On a typical Debian or Ubuntu system this may be sufficient:
+       apt-get install libsdl2-dev libsdl2-net-dev libsdl2-image-dev \
+         libpcre3-dev libsdl2-mixer-dev libfluidsynth-dev \
+         libportmidi-dev libmad0-dev libdumb1-dev libvorbis-dev
+         
+    You will also need `cmake` and `make` if you haven't already got them.
+         
+  2. `cd' to the directory containing the PrBoom distribution (the directory
+     this file is in, prboom2).
+
+     Run `cmake .` from this folder
+     
+     This should end with the message "Build files have been written to: [folderpath]"
+
+  3. Type `make' to compile PrBoom. This may take some time; while it's
+     compiling I suggest you read the README, or maybe go and look for some
+     good doom levels to play when it's finished :-).
+
+     This should work on all Linux systems, but systems that have a make
+     utility other than GNU make may have problems. E.g. you might need to
+     install GNU make, and then use it instead of make for these
+     instructions. I have tested BSD pmake and it works fine.
+
+  4. Type `make install-strip' as root, to install the programs, data files
+     and man pages for PrBoom. If you don't have root access on the machine,
+     you should ask the syadmin to do this and the next step for you.
+     
+     If you want to install manually then you must put the `prboom-plus.wad` 
+     in /usr/local/share/games/doom/ (or symlink it from there).
+
+  5. Copy your Doom, Doom 2, Ultimate Doom or Final Doom IWAD (doom.wad or
+     doom2.wad) to /usr/local/share/games/doom/ (or symlink it from there).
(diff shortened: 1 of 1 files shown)


prboom2/INSTALL
```diff
@@ -74,6 +51,9 @@ Basic Installation
      If you have a system with many users, you should read the license for
      your version of Doom, and make sure only those users allowed to use it
      can access the IWAD file.
+     
+     You can also aquire FreeDoom from https://freedoom.github.io/ which is a 
+     free replacement to the iD Doom iWADs (although if you intend to upload demos to DSDA this iWAD isn't sufficient)
```
kraflab, line 56: I think you can remove the `(although if you intend to upload demos to DSDA this iWAD isn't sufficient)` from here. That's certainly correct but it's probably not in the scope of the INSTALL doc :^)

prboom2/INSTALL
```diff
@@ -41,17 +17,15 @@ Basic Installation
...
+     this file is in, prboom2).
 
-     Type `./configure' to configure PrBoom for your system.  If you're
-     using `csh' on an old version of System V, you might need to type
-     `sh ./configure' instead to prevent `csh' from trying to execute
-     `configure' itself.
-
-     Running `configure' takes awhile.  While running, it prints some
-     messages telling which features it is checking for.
+     Run `cmake .` from this folder
```
jackrjli, line 26: I'd recommend adding the `-DCMAKE_BUILD_TYPE=Release` flag to get a release build (cmake defaults to debug, which results in lower performance).

prboom2/INSTALL
crabcrabcam: Sorted that, sorry. I forgot I wasn't committing this to the DSDA branch.

prboom2/INSTALL
crabcrabcam: I've made a change to this. I added a comment to say to remove it if you're wanting a debug build (I tend to run things as debug due to never quite being sure I've done everything right).

prboom2/INSTALL
```diff
@@ -41,17 +17,18 @@ Basic Installation
...
 
-     Type `./configure' to configure PrBoom for your system.  If you're
-     using `csh' on an old version of System V, you might need to type
-     `sh ./configure' instead to prevent `csh' from trying to execute
-     `configure' itself.
-
-     Running `configure' takes awhile.  While running, it prints some
-     messages telling which features it is checking for.
+     Run `cmake . -DCMAKE_BUILD_TYPE=Release` from this folder
+     (You should remove the -DCMAKE... if you wish to get a debug build
```
fabiangreffrath, line 27: More exactly, you should provide `-DCMAKE_BUILD_TYPE=Debug` for a debug build. I would never recommend anyone to invoke CMake without any `CMAKE_BUILD_TYPE` parameter at all.

prboom2/INSTALL
```diff
@@ -41,17 +17,18 @@ Basic Installation
...
+     for better info about crashes and such if using experimental code.
+     This does reduce performance)
```
fabiangreffrath, line 29: There is a stray closing parenthesis in this sentence.

prboom2/INSTALL
```diff
@@ -41,17 +17,18 @@ Basic Installation
...
+     
+     This should end with the message "Build files have been written to: [folderpath]"
```
fabiangreffrath, line 31: This is missing a full stop at the end. Please write complete sentences.

prboom2/INSTALL
```diff
@@ -74,6 +54,9 @@ Basic Installation
      If you have a system with many users, you should read the license for
      your version of Doom, and make sure only those users allowed to use it
      can access the IWAD file.
+     
+     You can also aquire FreeDoom from https://freedoom.github.io/ which is a 
+     free replacement to the iD Doom iWADs
```
fabiangreffrath, line 59: This is another incomplete sentence. Also, please prefer "may" over "can".

prboom2/INSTALL
```diff
@@ -65,6 +42,9 @@ Basic Installation
   4. Type `make install-strip' as root, to install the programs, data files
      and man pages for PrBoom. If you don't have root access on the machine,
      you should ask the syadmin to do this and the next step for you.
+     
+     If you want to install manually then you must put the `prboom-plus.wad`
```
fabiangreffrath, line 46: Please prefer "should" over "must".

prboom2/INSTALL
```diff
@@ -90,6 +73,33 @@ Basic Installation
      See the README file for information about getting the Timidity
      patches needed for music support, and more.
 
+
```
fabiangreffrath, line 76: Now this part is complete obsolete and irrelevant. We should not document building of the original project here which this fork is going to replace.

fabiangreffrath (changes requested):
Thank you very much for doing this! The documentation for this project surely needs some improvements. I request a few changes but are else fine with it.

alexey-lysiuk (commented):
I wouldn’t recommend an in-tree build. While it works without an issue at the moment, I suggest to create a new directory, and use it as a CMake build directory.

crabcrabcam:
> I wouldn’t recommend an in-tree build. While it works without an issue at the moment, I suggest to create a new directory, and use it as a CMake build directory.

I'm not sure how to make this one work. 

`mkdir build && cd build`
`cmake ..`
Then running `make` gives me `make: *** No targets specified and no makefile found.  Stop.` and running `make ..` gives me `make: Nothing to be done for '..'.` 

Can you give me a hand with that?

crabcrabcam:
@fabiangreffrath  I'm going to go through all the things you marked in this comment. 

I'm going to start with a dickish comment, please don't take it too personally. I'm not quite fond of you attacking my missed fullstops and telling me they are "incomplete sentences". I gather from your writing you are not a native English speaker, however your tone comes across as attacking someone who is trying to help. A missed fullstop is not something that requires a full comment on, but a note in another (or joining the two missed fullstop points) and perhaps instead of stating that it is "an incomplete sentence", you could word it that it's a typo, mistake, or just a small fix to make everything look slightly more professional. I say this not because I think you're wrong, but because if someone who is not perfect at English were to try to write some documentation I don't want them getting scared off by having all their grammatical mistakes pointed out in such a harsh tone. 

Other than that bit I agree with all your comments about the small changes. I can't see the missing parenthesis, but I've rewritten that sentence anyway as it was the one about debug builds. 

I've also removed the Timidity thing. I didn't know what it was, and reading the README wasn't going to hurt anyone, so I left it in. 

If you haven't seen my other comment about using a build folder, I could use some help with that as it doesn't appear to be working how I think it should (I'm not the best versed in make and C code, or code at all these days). 

Thanks for the help with this. I'm glad I could update the docs to be actually working after struggling through them to build this and failing.

fabiangreffrath:
@crabcrabcam I am sorry that you felt offended by my comments! Be certain that I did not mean to attack you in any way. You are right that I am not a native speaker either, so it's probably a bit hard for me to dose my tone. To me, a sentence with a missing full stop is not complete and thus, if I find one, I report an incomplete sentence. Absolutely no offense intended!

Regarding the build directory, you'll have to create it relative to the directory where the main 'CMakeList.txt` is located (i.e. `prboom2`) so that after `mkdir build && cd build` the next `cmake ..` points right there. Make sure to remove all remnants of previous cmake invocations before doing this, e.g. `CMakeCache.txt`, `CMakeFiles`, `cmake_install.cmake`, etc.

alexey-lysiuk:
> I'm not sure how to make this one work.
> 
> `mkdir build && cd build`
> `cmake ..`

If the current directory was the repository root initially, the second command should be `cmake ../prboom2`.

crabcrabcam:
> @crabcrabcam I am sorry that you felt offended by my comments! Be certain that I did not mean to attack you in any way. You are right that I am not a native speaker either, so it's probably a bit hard for me to dose my tone. To me, a sentence with a missing full stop is not complete and thus, if I find one, I report an incomplete sentence. Absolutely no offense intended!
> 
> Regarding the build directory, you'll have to create it relative to the directory where the main 'CMakeList.txt`is located (i.e.`prboom2`) so that after `mkdir build && cd build`the next`cmake ..`points right there. Make sure to remove all remnants of previous cmake invocations before doing this, e.g.`CMakeCache.txt`, `CMakeFiles`, `cmake_install.cmake`, etc.

Text is always tough with tone, sorry for the missunderstanding. I just don't want people with slightly bad English committing to a project and getting rejected for it. After all, poor English docs are better than docs that are outdated or missing. 

As for the inline build thing, deleting the cache and makefiles from the `prboom2` folder worked perfectly. It must have been seeing them and not running cmake on the main folder. Thanks for the help on that, I'll update the file now for that.

prboom2/INSTALL
fabiangreffrath: There is still a full stop missing after this sentence.

prboom2/INSTALL
fabiangreffrath: Please remove the paragraph about building from SVN.

fabiangreffrath (changes requested):

crabcrabcam:
There's a lot of references to the `configure` file, but I'm not sure what can be removed as most of it is also arguments for cmake. 

If someone wants that fixed can they please do it themselves as explaining to me what needs to be changed with that lot is going to be a lot harder than just doing it. 

Fingers crossed this is the last change needed and we can finally update the docs so people can compile the project :crossed_fingers:

PR #153, opened 2020-12-19
Save / load cl 9 friction

kraflab:
In cl 9, friction works via `friction_t` thinker objects spawned for each sector. Unfortunately, this type of thinker was never added to the save file logic. As a result, when playing in cl 9 specifically, if you reload a game on a map with special friction, the friction information is missing.

This PR adds support for the `friction_t` object in the save logic that handles the thinkers.

### DIFF
--- a/prboom2/src/p_saveg.c
+++ b/prboom2/src/p_saveg.c
@@ -559,7 +559,8 @@ enum {
   tc_scroll,      // killough 3/7/98: new scroll effect thinker
   tc_pusher,      // phares 3/22/98:  new push/pull effect thinker
   tc_flicker,     // killough 10/4/98
-  tc_endspecials
+  tc_endspecials,
+  tc_friction // store friction for cl 9
 } specials_e;
 
 //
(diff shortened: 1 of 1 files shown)


prboom2/src/p_saveg.c
```diff
@@ -559,7 +559,8 @@ enum {
   tc_scroll,      // killough 3/7/98: new scroll effect thinker
   tc_pusher,      // phares 3/22/98:  new push/pull effect thinker
   tc_flicker,     // killough 10/4/98
-  tc_endspecials
+  tc_endspecials,
+  tc_friction // store friction for cl 9
```
kraflab, line 563: We could add this before the `tc_endspecials` value, but that value is always used with an equality check, so it doesn't need to be at the end. If we changed the value of `tc_endspecials` then all old save files would be broken when loaded in this version (the end marker would be interpreted as friction instead and the load process would error out).

fabiangreffrath (approved):

PR #155, opened 2020-12-21
allow MBF sky transfers in all complevels

fabiangreffrath:


fabiangreffrath:
1. This makes the `-allow_sky_transfer_in_boom` obsolete (actually, it already is), so it should be removed.
2. This needs a compatibility flag (similar to e.g. "Sky is unaffected by invulnerability") for purists.

fabiangreffrath:
What's the general opinion on this one?

kraflab (approved):
Makes sense to me

fabiangreffrath:
One last question: Should this be enabled by default, as in e.g. Crispy, or disabled as in previous versions of PrBoom+?

fabiangreffrath:
Another last question: Native speakers, is it "in all complevels", "on all complevels" or "for all complevels"?

kraflab:
In this context, all three of those are correct, but the one I would use is "in". 🤠

fabiangreffrath:
> One last question: Should this be enabled by default, as in e.g. Crispy, or disabled as in previous versions of PrBoom+?

I guess I'll go with "disabled by default" to make this change less disruptive.

kraflab:
> > One last question: Should this be enabled by default, as in e.g. Crispy, or disabled as in previous versions of PrBoom+?
> 
> I guess I'll go with "disabled by default" to make this change less disruptive.

Makes sense to me as well

PR #157, opened 2021-01-02
fix psprite positioning problem for real

fabiangreffrath:
The FRACUNIT/4 value comes from Crispy Doom. It's a compromise between
the weapon sprite floating in the air (FRACUNIT/2) and a line of
Tutti-Frutti printed on top (0).


fabiangreffrath:
This needs some actual feedback from some specific users on the forums (e.g. Diabolución).

PR #158, opened 2021-01-02
make sure the canonical resolutions are always available

fabiangreffrath:
This adds the four resolutions known from Vanilla Doom, MBF and Doom
95, respectively, to the list of available resolutions, no matter
what.

Also, remove the unused screen_resolution_lowest variable, fix the
return type for the I_GetModeFromString() function, disable window
resizing in OpenGL mode and add a few comments to some non-obvious
code.

### DIFF
--- a/prboom2/src/SDL/i_video.c
+++ b/prboom2/src/SDL/i_video.c
@@ -133,7 +133,7 @@ int             leds_always_off = 0; // Expected by m_misc, not relevant
 extern int     usemouse;        // config file var
 static dboolean mouse_enabled; // usemouse, but can be overriden by -nomouse
 
-int I_GetModeFromString(const char *modestr);
+video_mode_t I_GetModeFromString(const char *modestr);
 
 /////////////////////////////////////////////////////////////////////////////////
 // Keyboard handling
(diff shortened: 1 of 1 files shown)


fabiangreffrath:
Could someone please review this and tell me if we actually want this? I find it quite useful, since my laptop screen claims to only support its own native resolution, and that's what I am left with in the Setup menu then.

prboom2/src/SDL/i_video.c
```diff
@@ -701,23 +711,29 @@ static void I_FillScreenResolutionsList(void)
...
+    count = MIN(count, MAX_RESOLUTIONS_COUNT - 2 - num_canonicals);
 
-    for(i = count - 1; i >= 0; i--)
+    for(i = count - 1 + num_canonicals; i >= 0; i--)
     {
       int in_list = false;
 
-      SDL_GetDisplayMode(display_index, i, &mode);
+      // make sure the canonical resolutions are always available
+      if (i > count - 1)
```
kraflab, line 725: Is it possible for the canonical resolutions to already be included in the display modes returned by SDL? Would we end up with duplicate entries in the list?

fabiangreffrath: Nope, that's accounted for a few lines down (check for the `in_list` variable). SDL would return all the different *video modes* that your display supports - i.e. all possible combinations of resolution, color depth, refresh rate, etc. - whereas we only take the resolution into consideration. Thus, we already expect the actual resolutions list to be made up of a list of video modes that may contain duplicate resolutions.

kraflab (approved):

PR #160, opened 2021-01-03
get rid of the "Screen Multiple Factor" feature

fabiangreffrath:
Since the switch to SDL2 we have seamless scaling of the game window.
This means, we don't need to explicitly multiply pixels in order to
scale a low resolution up to the desired window size anymore. This
works for all rendering modes, both windowed and fullscreen - except
for OpenGL rendering, but this doesn't support the "screen multiple
factor" feature either.


fabiangreffrath:
In Chocolate Doom and related ports, the `intermediate_texture` upscales the framebuffer by means of nearest scaling to the next integer multiple of the original resolution which exceeds the screen dimensions. The picture is then downscaled to the screen by means of linear scaling.

Here, however, the `upscaled_texture` seems to provide a mere integer upscaling of the original framebuffer without taking the screen size into account or making use of different scaling methods.

fabiangreffrath:
I intend to replace this feature and its menu entry with an "integer scaling" option.

PR #162, opened 2021-01-04
fix aspect ration for the canonical screen resolutions

fabiangreffrath:


fabiangreffrath:
Bildschirmfoto von 2021-01-04 22-32-36

fabiangreffrath:
Still need to check if the fix is actually necessary for 640x400 or if a similar result could be achieved with a specific "Status Bar and Menu Appearance" setting.

PR #163, opened 2021-01-05
properly reset software renderer to 8bit when left with OpenGL

fabiangreffrath:


fabiangreffrath (commented):
Checked that `V_InitMode()` is only called from this one location, so this change should be safe.

PR #168, opened 2021-01-15
back to one binary "prboom-plus" with optional OpenGL rendering

### LINKED ISSUE #166: back to one binary "prboom-plus" with OpenGL rendering
The current situation with two binaries - one with support for OpenGL rendering and one without - is a relic of the 1990. I don't remember which version of Windows 95/98 exactly came with OpenGL out-of-the-box, but it was not a given fact back then that your OS would support that interface. However, today every mobile phone could be expected to provide such functionality.

So, I propose to build only one binary by default which has of course OpenGL enabled. We should leave it as a compile time option, though, for those rare cases that don't support it, but by today's standards it should be enabled by default. Thus, this would merely boil down to a change in the CMake build system, all the `#ifdef GL_DOOM` checks in the code should remain in place.

@alexey-lysiuk As you are the main author of our current CMake build system, would you like to have a look at this?

fabiangreffrath:
Fixes #166.

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -94,11 +94,11 @@ jobs:
         cd build
         mkdir package
         if [[ "${{ runner.os }}" == 'Windows' ]]; then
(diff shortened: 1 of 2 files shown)


kraflab:
Nice 👍

PR #172, opened 2021-01-18
if a map is started during demo recording, start a new demo

### LINKED ISSUE #88: [Help required] Trying to implement demo restart key, but hitting wall
I've recently been made aware of Crispy's feature of being able to restart demo recordings with a keypress (the "Restart current map" key), first introduced [here](https://github.com/fabiangreffrath/crispy-doom/commit/2b214e313f409ee153941d749ba2604e281e265d). 

I've been attempting to introduce the same feature into prboom+ basing myself on Crispy's implementation, but I'm running into a particular crash and also parts of the code I'm not sure I fully understand. These are my commits so far into a new branch of my [fork](https://github.com/physixcat/prboom-plus/commit/209663997bf8f95e7c1134f3b7551bf4b8b7610d)

When you're recording a demo, and the restart demo key is pressed, the program crashes in the following fashion:
```
P_GetNodesVersion: using normal BSP nodes
gld_Precache: MAP01 done in 71 ms
G_CheckDemoStatus: Demo recorded
P_GetNodesVersion: using normal BSP nodes
Z_Free: freed a pointer without ZONEID
I_ShutdownMusic: removed /tmp/prboom-plus-music-COiMTN
I_ShutdownSound:
```
note the **Z_Free: freed a pointer without ZONEID** . I'm also uncertain what the demowarp variable from crispy does, and if it has anything to do with this problem since I haven't introduced it yet.

Can't seem to figure out how to fix this. I was hoping maybe @fabiangreffrath could enlighten me on how best to approach the problem of implementing demo restarts and also how to fix that crash. Looking forward to your replies, trying to learn as I go here.

fabiangreffrath:
Fixes #88.

### DIFF
--- a/prboom2/src/e6y.c
+++ b/prboom2/src/e6y.c
@@ -376,9 +376,15 @@ int G_ReloadLevel(void)
 
   if ((gamestate == GS_LEVEL) &&
       !deathmatch && !netgame &&
-      !demorecording && !demoplayback &&
+      !demoplayback &&
       !menuactive)
   {
+    // restart demos from the map they were started
+    if (demorecording)
+    {
+      gameepisode = startepisode;
+      gamemap = startmap;
+    }
     G_DeferedInitNew(gameskill, gameepisode, gamemap);
     result = true;
   }
--- a/prboom2/src/e6y.c
+++ b/prboom2/src/e6y.c
@@ -376,7 +376,7 @@ int G_ReloadLevel(void)
 
   if ((gamestate == GS_LEVEL) &&
       !deathmatch && !netgame &&
-      !demoplayback &&
+      !democontinue && !demoplayback &&
       !menuactive)
   {
     // restart demos from the map they were started
--- a/prboom2/src/g_game.c
+++ b/prboom2/src/g_game.c
@@ -2421,12 +2421,51 @@ static skill_t d_skill;
 static int     d_episode;
 static int     d_map;
 
+static char *G_NewDemoName(const char *name)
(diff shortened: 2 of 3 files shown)


fabiangreffrath:
@kraflab I based this on your implementation in dsda-doom, which in turn is based on the one from Crispy Doom,

Am I missing something?

- [ ] How do we cover the `democontinue` case?
- [ ] You reset the `basetic` variable in `G_DeferedInitNew()`. Doesn't this function mean that a new game is started only in the *next* tic?
- [ ] Why do you call `G_RecordDemo()` from `G_DeferedInitNew()` but `G_BeginRecording()` only from `G_DoNewGame()`?

Thanks!

kraflab:
I don't see anything missing.

- `democontinue`: are you asking what to do in the case where a user joins in the middle of demo playback and then presses the button? I didn't do anything special for this case in dsda-doom because it's superseded by the more robust [key frame](https://github.com/kraflab/dsda-doom/blob/master/prboom2/src/dsda/key_frame.c) system that's currently in development.
- `basetic`: it works in this position - it could probably also be set elsewhere
- I believe `G_RecordDemo` and `G_BeginRecording` were placed to mirror the arrangement of these calls when the program starts. `G_RecordDemo` turns `demorecording` back on (after being shut off by `G_CheckDemoStatus`), and `G_BeginRecording` can't be run until later on as far as I know.

fabiangreffrath:
Thank you for your review and explanations!

PR #178, opened 2021-01-24
Updates to the UMAPINFO docs to disambiguate the 'bossaction' field a bit

XaserAcheron:
Just a quick PR that edits umapinfo.txt to clear up some ambiguity around the 'bossaction' field. Quick summary of the updates:
- Added documentation for `bossaction = clear`
- Clarified that defining a UMAPINFO `bossaction` will replace any existing boss death actions (e.g. E1M8 barons, E2M8 cybs...), rather than add to it.
- Added list of valid thingtypes at the bottom of the doc (bossaction's comment has a "see below" comment, but there was nothing below ;)

At some point it might be good to formalize this a bit more as a proper spec, rather than rely on the source as a reference implementation, but I'm definitely not the person to lead that charge. :P


fabiangreffrath:
Great, thank you!

PR #179, opened 2021-01-24
Externalize KFA cheat string. Resolves #177.

### LINKED ISSUE #177: STSTR_KFAADDED isn't read from BEX patches
See title. Seems to be a very old bug (same issue with Boom).

JadingTsunami:
Use the externalized version of the KFA cheat string. Resolves issue #177.


fabiangreffrath:
Thank you!

PR #180, opened 2021-01-26
support named complevels on the command line, e.g. "-complevel boom"

fabiangreffrath:
The names are based on the args to Chocolate Doom's "-gameversion"
parameter and are currently restricted to the most usual 5 complevels.

### DIFF
--- a/prboom2/src/g_game.c
+++ b/prboom2/src/g_game.c
@@ -2006,18 +2006,23 @@ unsigned int GetPackageVersion(void)
 
 // [FG] support named complevels on the command line, e.g. "-complevel boom",
 // names based on the args to Chocolate Doom's "-gameversion" parameter
+// and IWAD file names
 
 static int GetComplevel (const char *arg)
 {
   int i;
 
-  struct {
-    const int level;
+  const struct {
+    int level;
     const char *const name;
   } named_complevel[] = {
     {2, "1.9"},
+    {2, "doom2"},
     {3, "ultimate"},
+    {3, "udoom"},
     {4, "final"},
+    {4, "tnt"},
+    {4, "plutonia"},
     {9, "boom"},
     {11, "mbf"},
   };
--- a/prboom2/src/g_game.c
+++ b/prboom2/src/g_game.c
@@ -2031,7 +2031,7 @@ int G_GetNamedComplevel (const char *arg)
   // choose the complevel based on the IWAD
   if (!strcasecmp(arg, "vanilla"))
   {
-    if (gamemode == retail)
+    if (gamemode == retail || gamemission == chex)
       return 3;
     else if (gamemode == commercial && (gamemission == pack_plut || gamemission == pack_tnt))
       return 4;
--- a/prboom2/src/g_game.c
+++ b/prboom2/src/g_game.c
@@ -2028,6 +2028,7 @@ int G_GetNamedComplevel (const char *arg)
     {11, "mbf"},
   };
 
+  // choose the complevel based on the IWAD
   if (!strcasecmp(arg, "vanilla"))
   {
     if (gamemode == retail)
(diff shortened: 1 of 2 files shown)


prboom2/src/g_game.c
```diff
@@ -2004,6 +2004,35 @@ unsigned int GetPackageVersion(void)
...
+
+static int GetComplevel (const char *arg)
+{
+  int i;
+
+  struct {
+    const int level;
+    const char *const name;
+  } named_complevel[] = {
+    {2, "1.9"},
```
kraflab, line 2018: Would "doom2" be better here?

prboom2/src/g_game.c
```diff
@@ -2004,6 +2004,35 @@ unsigned int GetPackageVersion(void)
...
+    {3, "ultimate"},
+    {4, "final"},
```
kraflab, line 2020: We could have extra aliases here as well, like "tnt" and "plutonia". Another would be "doom" for cl 3. That way you can basically pass the iwad name for the complevel.

fabiangreffrath: Very good idea! I'll add the IWAD names as aliases. I am a bit unsure about plain "doom", though, as that's a bit too ambiguous for my taste. The original non-Ultimate Doom had the "doom2" executable after all.

kraflab: It could be ambiguous, I'm not sure. I'd probably still use it but i don't have a super strong opinion on it.

kraflab: If you want to be even more aggressive on this feature we could just have vanilla, boom, and mbf, where vanilla chooses the complevel based on the iwad.

fabiangreffrath: > If we want to be even more aggressive on this feature we could just have vanilla, boom, and mbf, where vanilla chooses the complevel based on the iwad.

But isn't that the standard behavior anyway?

fabiangreffrath: > But isn't that the standard behavior anyway?

For demos, yes, but not for regular game play.

prboom2/src/g_game.c
```diff
@@ -2004,6 +2004,52 @@ unsigned int GetPackageVersion(void)
...
+    {4, "tnt"},
+    {4, "plutonia"},
+    {9, "boom"},
+    {11, "mbf"},
+  };
+
+  // choose the complevel based on the IWAD
+  if (!strcasecmp(arg, "vanilla"))
+  {
+    if (gamemode == retail)
```
kraflab, line 2034: If we want to account for Chex, it has the `gamemode` set to `shareware` since it's one episode, but it's based on ultimate doom so the correct complevel is 3 :^)

fabiangreffrath: According to Doomwiki, chex.exe is a modified version of Final Doom's executable.

kraflab: Interesting, it's a modified ultimate doom wad but final doom exe 😮

Hm, for consistency with other doom 1 style wads we've always recorded in cl 3. 🤔

fabiangreffrath: Whatever, I just checked and current PrBoom+ plays Chex's IWAD demos with Ultimate Doom compatibility.

As far as I know, the only relevant change between the Ultimate and the Final Doom executable was the teleporter z-coordinate bug. However, there was a second Final Doom executable released in 1996, i.e. the same year as Chex, which had this bug fixed. And even Chocolate Doom, i.e. the source port that first introduced specific support for Chex, only checks for the original Final Doom executable to reproduce this bug:

https://github.com/chocolate-doom/chocolate-doom/blob/b26157aca5b12049b35d8dfcf969c51967a369f6/src/doom/p_telept.c#L99

In short terms, I guess it's reasonable and safe to assume Ultimate Doom compatibility for Chex. :wink:

fabiangreffrath:
Alright, I am fine with this PR now. Any objections left?

kraflab (approved):

PR #182, opened 2021-01-28
add support for colored blood and gibs

fabiangreffrath:
Due to popular demand. :wink:

### DIFF
--- a/prboom2/src/p_map.c
+++ b/prboom2/src/p_map.c
@@ -39,7 +39,6 @@
 #include "p_map.h"
 #include "p_setup.h"
 #include "p_spec.h"
-#include "p_tick.h" // P_SetTarget()
 #include "s_sound.h"
 #include "sounds.h"
 #include "p_inter.h"
@@ -1972,7 +1971,7 @@ dboolean PIT_ChangeSector (mobj_t* thing)
     }
     thing->height = 0;
     thing->radius = 0;
-    P_SetTarget(&thing->target, thing);
+    thing->flags |= P_ColoredBlood(thing);
     return true; // keep checking
     }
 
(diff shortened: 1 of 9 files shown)


kraflab:
On the speedrunning side, we actually don't allow any kind of sprite changes like this. Obviously prboom+ isn't constrained to dsda rules, which is why there's a separate port in the first place, but I felt like I should mention it anyway for posterity.

prboom2/src/p_map.c
```diff
@@ -2008,6 +2009,7 @@ dboolean PIT_ChangeSector (mobj_t* thing)
     mo = P_SpawnMobj (thing->x,
                       thing->y,
                       thing->z + thing->height/2, MT_BLOOD);
+    mo->target = thing;
```
kraflab, line 2012: With the way prboom+'s pointer cleanup works, you should use P_SetTarget instead of editing the target value directly. It probably wouldn't happen under normal constraints that a monster hit with an attack might be removed from memory, but I bet it's possible with dehacked, which could lead to some odd behaviour.

coelckers: Yes, it is indeed possible - and actually, doing this unconditionally may cause demo desyncs - not with the regular blood setup but if you add some code pointers that may work with the target pointer, all bets will be off.
I'd actually try to store the MT_ value somewhere instead of a target reference which may cause unforeseen issues

fabiangreffrath: Whoa, I wasn't even aware of `P_SetTarget()`, thanks!

> I'd actually try to store the MT_ value somewhere instead of a target reference which may cause unforeseen issues

I'll see what I can do, but if I can't misuse an existing field in the `mobj_t` struct, then I'll have to introduce something new anywhere else. Stay tuned...

fabiangreffrath:
So, what do you guys say?

kraflab (approved):

PR #183, opened 2021-01-31
Fix odd strafe values

kraflab:
There's a major bug in prboom+ that has been there since at least the beginning of the commit history stored in git (>20 years).

In vanilla doom, mouse strafe only occurs in multiples of 2, but in prboom+ you can achieve any value. This is not particularly relevant for casual players, but being able to produce values of 1 has a pretty big impact on gliding with mouse strafe (for good or bad depending on the situation).

I don't know exactly when this behaviour changed, but the original is still there in the woof source, so in principal this fix should apply to cl 9 and cl 11 for instance as well. However, I've only changed it for pre-boom complevels. If this kind of behaviour is "wrong" in the port most commonly used for those complevels for recording, for 20 years, it's probably not something we should change. 🤷

### DIFF
--- a/prboom2/src/g_game.c
+++ b/prboom2/src/g_game.c
@@ -700,7 +700,15 @@ void G_BuildTiccmd(ticcmd_t* cmd)
     forward += mousey;
   }
   if (strafe)
+  {
     side += mousex / movement_mousestrafedivisor; /* mead  Don't want to strafe as fast as turns.*/
+
+    // You can only produce even strafe values in vanilla doom
+    if (demo_compatibility)
+    {
+      side &= ~1;
+    }
+  }
   else
     cmd->angleturn -= mousex; /* mead now have enough dynamic range 2-10-00 */
 
--- a/prboom2/src/g_game.c
+++ b/prboom2/src/g_game.c
@@ -702,12 +702,7 @@ void G_BuildTiccmd(ticcmd_t* cmd)
   if (strafe)
   {
     side += mousex / movement_mousestrafedivisor; /* mead  Don't want to strafe as fast as turns.*/
-
-    // You can only produce even strafe values in vanilla doom
-    if (demo_compatibility)
-    {
-      side &= ~1;
-    }
+    side = (side / 2) * 2; // only even values are possible
   }
   else
     cmd->angleturn -= mousex; /* mead now have enough dynamic range 2-10-00 */


fabiangreffrath:
What's the situation in Choco?

kraflab:
Choco and crispy both have only even values, so they are correct.

At some point mead (based on the comment in the code) updated the mouse code to make mouse strafe slower for convenience. But where crispy did this by adding a new scale, mead did this by changing the factor that multiples mousex when calculating side move from `* 2` to `/ 4`, which is how odd values became possible.

fabiangreffrath:
Apart from strictly being correct, what other consequences does this have?

kraflab:
I don't think this really impacts anything besides gliding.

fabiangreffrath:
Okay, I am just asking because you called it a "major bug". :wink: So, in the end this makes it easier to apply fine-grained strafe movement than should be possible with any of the DOS executables, right?

kraflab:
I would say anything that allows you to produce inputs that are supposed to be impossible is a major bug 😸

It's basically like prboom+ has longtics for mouse strafe right now.

fabiangreffrath:
I see. My instant reaction would be to re-introduce an even factor to multiply with `mousex`, but now we have already introduced `movement_mousestrafedivisor` which doesn't make things easier. Erasing out the lowest bet does indeed seem to be the most pragmatic solution.

fabiangreffrath:
*bit*

fabiangreffrath:
Hm, I am still not entirely convinced by this change. First, because it erases a bit where there was a multiplication before (i.e. that's two very different methods of constructing multiples of two). Second, because it only applies to Vanilla complevels, which seems like a rather arbitrary restriction, given that the code has remained unchanged from Vanilla to at least MBF and was only changed in PrBoom (probably still without "plus" at that time).

prboom2/src/g_game.c
```diff
@@ -700,7 +700,15 @@ void G_BuildTiccmd(ticcmd_t* cmd)
     forward += mousey;
   }
   if (strafe)
+  {
     side += mousex / movement_mousestrafedivisor; /* mead  Don't want to strafe as fast as turns.*/
+
+    // You can only produce even strafe values in vanilla doom
+    if (demo_compatibility)
+    {
+      side &= ~1;
```
kraflab, line 709: Is this right for negative side values? 🤔 
I guess that would actually turn -1 into -2? :^)

kraflab:
I mean, we can also divide by 2 and then multiply by 2 😅 
We should preserve the magnitudes so that it doesn't make everyone have to adjust their sensitivity.

The thing is that this is a pretty important difference. It's not competitively fair for crispy and prboom+ to have different values possible for this input, so the most important thing is to fix the behaviour for the shared compatibilities.

For boom+, ya idk 🤷 
Maybe we should just fix it for everything.

We will probably ask any runners that use mouse strafe and prboom+ to upgrade unfortunately 😕

fabiangreffrath:
> I mean, we can also divide by 2 and then multiply by 2 sweat_smile

It makes more sense like this. Okay to merge?

kraflab:
> > I mean, we can also divide by 2 and then multiply by 2 sweat_smile
> 
> It makes more sense like this. Okay to merge?

Yep 🚀

fabiangreffrath:
Thank you!

PR #190, opened 2021-02-10
draw crosshair for all weapons all the time

D3Archi:
for some reason crosshair is not being drawn for melee weapons, it makes sense for vanilla, but custom wads may have different weapons on their place
crosshair also was not being drawn during messages like "secret is revealed" for some reason

### DIFF
--- a/prboom2/src/hu_stuff.c
+++ b/prboom2/src/hu_stuff.c
@@ -2262,10 +2262,8 @@ void HU_draw_crosshair(void)
 
   crosshair.target_sprite = -1;
 
-  if (!crosshair_nam[hudadd_crosshair] || crosshair.lump == -1 ||
-    custom_message_p->ticks > 0 || automapmode & am_active ||
-    menuactive != mnact_inactive || paused ||
-    plr->readyweapon == wp_chainsaw || plr->readyweapon == wp_fist)
+  if (!crosshair_nam[hudadd_crosshair] || crosshair.lump == -1 || automapmode & am_active ||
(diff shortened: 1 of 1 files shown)


fabiangreffrath:
I don't see the problem this is going to solve. Which melee weapon replacement could be better used with a crosshair?

D3Archi:
I've made a rifle replacement for chainsaw in this wad, for example:
https://www.doomworld.com/forum/topic/100425/

fabiangreffrath:
I see. Does it help if we change the weapon check to `weaponinfo[plr->readyweapon].ammo == am_noammo`?

D3Archi:
Yeah, it's possible to change to that, but I don't really see why won't we just draw the crosshair for melee weapons as well? They are hitscan attacks and knowing the exact location could help too.

D3Archi:
Here's also the reason to move crosshair drawing code before messages.

Without fix:
image

With fix:
image

fabiangreffrath:
I buy that second change, but I am still not convinced of rendering a crosshair for a fist.

D3Archi:
Another wad with ranged weapon on place of chainsaw, this time it has infinite ammo:
https://dsdarchive.com/wads/sf2012

kraflab:
I don't really see a reason not to just draw the crosshair for everything tbh. Even for the fist it would be useful for, e.g., punching things like mancubi. I don't use the crosshair though. Would be curious if any users object.

fabiangreffrath:
Okay, now you got me. 😁

PR #196, opened 2021-02-12
preserve complevel after using cheat for the rest of the session

D3Archi:
Right now it gets replaced with default value from config each time you reset level or change level

### DIFF
--- a/prboom2/src/doomstat.c
+++ b/prboom2/src/doomstat.c
@@ -50,7 +50,7 @@ dboolean modifiedgame;
 //-----------------------------------------------------------------------------
 
 // CPhipps - compatibility vars
-complevel_t compatibility_level, default_compatibility_level;
+complevel_t compatibility_level, default_compatibility_level, last_compatibility_level;
 
 // e6y
 // it's required for demos recorded in "demo compatibility" mode by boom201 for example
--- a/prboom2/src/doomstat.h
+++ b/prboom2/src/doomstat.h
@@ -68,7 +68,7 @@ extern const char *doomverstr;
 extern  dboolean modifiedgame;
 
 // CPhipps - new compatibility handling
-extern complevel_t compatibility_level, default_compatibility_level;
+extern complevel_t compatibility_level, default_compatibility_level, last_compatibility_level;
 
 // CPhipps - old compatibility testing flags aliased to new handling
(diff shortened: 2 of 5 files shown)


fabiangreffrath (approved):
LGTM

fabiangreffrath:
So, this preserves complevels changes in-game across level starts. How about the complevel set on the command line, which one gets precedence?

D3Archi:
For single player: cheat > command-line parameter > cfg value.
For demo playback cheat is not accessible, old rules apply.

fabiangreffrath:
Sounds reasonable. @kraflab ?

kraflab:
> For single player: cheat > command-line parameter > cfg value.

It looks to me like command-line parameter wins. Am I missing something?
```c
  compatibility_level = last_compatibility_level;
  {
    int i = M_CheckParm("-complevel");
    if (i && (1+i) < myargc) {
      int l = G_GetNamedComplevel(myargv[i+1]);;
      if (l >= -1) compatibility_level = l;
    }
  }
```

D3Archi:
That is right, I completely missed that.

Separating default config from current session config would help a lot with all those extra CheckParam calls I believe.

fabiangreffrath:
So, how do we proceed with this?

D3Archi:
I want to make a dedicated structure for settings that are only changeable during session, like complevel, shorttics and most other stuff that is applied via command line parameters, so that it doesn't interfere with default settings and game wouldn't have to decide which setting to load, the one from command like or one from default values. Not a big fan of code calling M_CheckParm in runtime.

fabiangreffrath:
This is a bit dirty. But since the complevel set by cheating can only get overridden by the one set on the command line, you could check after cheating if the command line contains the `-complevel` string and modify the `argv[]` array so that the next element points to a char array that overrides the actual command line choice.

fabiangreffrath:
Probably like this:
```patch
static void force_set_complevel (void)
{
  int i;
  static char str[4] = "";

  while ((i = M_CheckParm("-complevel")) == 0)
    M_AddParam("-complevel");

  snprintf(str, sizeof(str), "%d", complevel);

  if ((i+1) < myargc)
    myargv[i+1] = str;
  else
    M_AddParam(str);
}
```

kraflab:

PR #198, opened 2021-02-12
allow to bind weapon selection to mouse buttons

### LINKED ISSUE #129: Cannot bind weapon to mouse-button through main menu
The title says it all. Mouse buttons such as mouse5 and 6 do not register on the menu which you rebind your weapons on.

fabiangreffrath:
Fixes #129.


D3Archi:
How about making an additional option in menu to force pistol start every time as well by default?

prboom2/src/m_menu.c
```diff
@@ -1985,8 +1985,7 @@ static void M_DrawSetting(const setup_menu_t* s)
 
     if (key) {
       M_GetKeyString(*key,0); // string to display
-      if (key == &key_up || key == &key_down || key == &key_speed ||
-         key == &key_fire || key == &key_strafe || key == &key_strafeleft || key == &key_straferight || key == &key_use)
+      if (s->m_mouse || s->m_joy)
```
kraflab, line 1988: Incidentally, I removed this branch condition entirely in dsda-doom because it doesn't seem to serve any purpose https://github.com/kraflab/dsda-doom/commit/8c8dadb6247e0690d0a35d0d43db6b79f9c47efb

prboom2/src/m_misc.c
```diff
@@ -736,6 +736,16 @@ default_t defaults[] =
   {"key_prevweapon",  {&key_prevweapon},      {KEYD_MWHEELDOWN},
    0,MAX_KEY,def_key,ss_keys}, // key to cycle to the previous weapon
 
+  {"mb_weapon1",{&mb_weapon1},{-1},-1,MAX_MOUSEB,def_int,ss_keys}, // mouse button to switch to weapon 1 (fist/chainsaw)
```
kraflab, line 739: This might belong at line 581 in the mouse bindings section of the config file.

kraflab:
I'm going to experiment with some ideas I have about how to make input configuration much easier in the code and also for users, so we don't always need to do these kinds of patches 😸

kraflab:
> I'm going to experiment with some ideas I have about how to make input configuration much easier in the code and also for users, so we don't always need to do these kinds of patches 😸

https://github.com/kraflab/dsda-doom/commit/03ee01ae0f4eefb363328f519f14c276a8443fdc?w=1

This is conceptually where I will probably go in dsda-doom (just migrated the forward input and set up the code to make it compatible with the menu / config file system). It will look cleaner when all the S_KEY values are migrated. That said it will be a lot of work to refactor everything and also not introduce any bugs 😸

fabiangreffrath:
Your approach is looking really good and I agree that it should have been implemented like this from the very beginning. However, the code we are looking at here is 20+ years old, and back then  mice had 3 buttons and joysticks maybe 4. It will be a huge amount of work to convert every binding to the new concept and, tbh, I am not sure if this is applicable to this port. Because, you know, dare to introduce one bug. For the time being,  I am fine to add mouse bindings for the most requested actions and these seem to be weapon choices currently.

kraflab:
> Your approach is looking really good and I agree that it should have been implemented like this from the very beginning. However, the code we are looking at here is 20+ years old, and back then mice had 3 buttons and joysticks maybe 4. It will be a huge amount of work to convert every binding to the new concept and, tbh, I am not sure if this is applicable to this port. Because, you know, dare to introduce one bug. For the time being, I am fine to add mouse bindings for the most requested actions and these seem to be weapon choices currently.

Totally agree 👍

kraflab:
https://github.com/kraflab/dsda-doom/pull/8 posting for posterity

PR #199, opened 2021-02-12
add a "-pistolstart" command line parameter

### LINKED ISSUE #132: Way to automatically "pistol start" every level (feature)
**Expected:**
Convenient way to "pistol start" every level.
In this "mode" the player starts each and every level with:
* 50 bullets and no other ammo
* a pistol, a fist and no other weapon
* 100% health and 0% armor

Player starts each and every level as if they just died and just respawned - automatically.

**Actual:**
Player has to remember to type IDCLEV... or to press Home when each new level starts after finishing the previous level.

**Possible solution:**
`-pistolstart` command line switch [like in crispy-doom](https://github.com/fabiangreffrath/crispy-doom/pull/659)

**Extra:**
It's not a huge deal I suppose as the IDCLEV... way is the way many/most of us always done "pistol start" all along I'd imagine but it would be nice if this could be automated/forced. I sometimes forget to press Home on a new level and then realise it later that I should restart the level as it wasn't a pistol start. In crispy-doom the command line switch works really conveniently.

fabiangreffrath:
Fixes #132.

### DIFF
--- a/prboom2/doc/prboom-plus-usage.txt
+++ b/prboom2/doc/prboom-plus-usage.txt
@@ -188,6 +188,7 @@ Additional command-line parameters:
 "-nowindow" - for temporarily using fullscreen mode without saving this in the cfg.
 "-aspect NxM" - for using a different aspect ratio; e.g. -aspect 5x4, -aspect 8x5 or -aspect 2x1.
 "-videodriver name" - for setting up the videodriver name that SDL will use. "windib" and "directx" are available for Windows. See SDL documentation for other platforms. "-videodriver default" can be used to force SDL behaviour by default.
+"-pistolstart" - automatic pistol start when advancing from one level to the next
 "-shorttics" - forces the same mouse behaviour as when recording (i.e. the converse of "-longtics")
 "-stroller" - removes ability to strafe and limits running speed (a speedrun category)
 "-resetgamma" - restores the original gamma after a crash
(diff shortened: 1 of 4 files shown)


D3Archi:
How about making an additional menu option to force pistol start by default as well?

fabiangreffrath:
I want to avoid adding a menu switch that breaks with regular gameplay in such a drastic way and is incompatible with anything else that this port focuses on.

kraflab:
Since we've been adding cheats, what do you think about making this a cheat instead of a command-line option? 🤔 
Then it doesn't work while recording anyway, and people can toggle it while in-game (and for the non-command-line-inclined). Just an idea 🤷

D3Archi:
I don't really think cheat suits well here. Cheats are situational, tntcl would be useful for those who launch different cl2/9/11 wads without command line parameters, and shorttics are only useful in some cases as well.

Forcing player start was already implemented in prboom+ good enough, pressing "home" button starts the level from player start. But since a new command line parameter is being introduced I think it would be fair to add an option in menu to enable player starts by default. For example, default complevel and shorttics are already in cfg.

kraflab:
I guess there are people who mostly play with pistol start only, so maybe it makes sense to be in the menu. That said, I'm not sure I would do anything personally - as archi mentioned, the restart map key already does well enough when people want to pistol start.

fabiangreffrath:
I have just merged this, so it is at least *possible* to have pistol starts. The decision whether there will be a menu switch for this or not is orthogonal to this.

PR #201, opened 2021-02-12
Fix typo in ZDoom-style Actor name.

JadingTsunami:
Recommend to fix this before any mods use the typo'd actor name and it splits the naming.

### DIFF
--- a/prboom2/doc/umapinfo.txt
+++ b/prboom2/doc/umapinfo.txt
@@ -164,7 +164,7 @@ RedTorch
 ShortBlueTorch
 ShortGreenTorch
 ShortRedTorch
-Slalagtite
+Stalagtite
 TechPillar
 CandleStick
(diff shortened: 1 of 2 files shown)


fabiangreffrath:
This is still a germanism. The proper English term is stalactite.

JadingTsunami:
> This is still a germanism. The proper English term is stalactite.

In fact, in this case stalagmite is the right word because the sprite rises up from the ground, but yes-- a typo no matter how you slice it. :)

However my understanding of the intent is to align with the ZDoom actor names, which (for better or worse) use Stalagtite. PrBoom could deviate in this one instance from the ZDoom Actor naming, but I am not sure I see why if all other actor names are aligned to the ZDoom de facto standard.

fabiangreffrath:
Indeed, thanks for bringing this up!

PR #211, opened 2021-02-21
Allow longtics when recording demos by using additional named complevels

ceski-1:
A new demo version (204) is used to differentiate complevel 9 demos with longtics enabled. A similar case is already allowed with "-complevel 2" which uses demo version 111 when longtics is enabled. If demo version 204 is not ideal, a different number can be used instead.

### DIFF
--- a/prboom2/src/g_game.c
+++ b/prboom2/src/g_game.c
@@ -2064,10 +2064,19 @@ int G_GetNamedComplevel (const char *arg)
     {4, "plutonia"},
     {9, "boom"},
     {11, "mbf"},
+    {2, "1.9_longtics"},
+    {2, "doom2_longtics"},
+    {3, "ultimate_longtics"},
+    {3, "udoom_longtics"},
+    {4, "final_longtics"},
+    {4, "tnt_longtics"},
+    {4, "plutonia_longtics"},
+    {9, "boom_longtics"},
+    {11, "mbf_longtics"},
   };
 
   // choose the complevel based on the IWAD
-  if (!strcasecmp(arg, "vanilla"))
+  if (!strcasecmp(arg, "vanilla") || !strcasecmp(arg, "vanilla_longtics"))
   {
     if (gamemode == retail || gamemission == chex)
       return 3;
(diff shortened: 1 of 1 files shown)


kraflab:
Absolutely not. If anything longtics should be removed from recording on lower complevels, not the reverse.

ceski-1:
Maybe as an alternative, always switch over to the latest PrBoom demo version (currently 214) when specifying "-longtics". The idea is to retain precise mouse movement when recording demos while still having the convenience of using the various complevel command-line parameters. Existing demos will still play correctly. What do you think?

kraflab:
I don't think it would be a good idea to record at a complevel different from the one specified on the command line.

ceski-1:
| named complevel | demo version | complevel (internal) |
| --- | --- | --- |
| 1.9_longtics<br>doom2_longtics | 111 | 2 |
| ultimate_longtics<br>udoom_longtics | 111 | 3 |
| final_longtics<br>tnt_longtics<br>plutonia_longtics | 111 | 4 |
| boom_longtics | 204 | 9 |
| mbf_longtics | 205 | 11 |

ceski-1:
Okay, I updated the code changes to reflect the new solution. The complevels are implemented as the "named" variations shown in my last comment. As expected, Boom and MBF demos with longtics enabled won't play on previous versions of PrBoom-Plus and will exit with an error thanks to the new demo versions in the file header.

The behavior of the "-longtics" parameter is unchanged. It only has an effect when used in combination with lower complevels, just like before. In fact, all previous methods of demo recording and playback are unchanged. The code changes simply provide extra options for anyone looking to record with more precise mouse movement.

Hopefully, by requiring the user to explicitly use something like "-complevel boom_longtics -record mydemo" when launching the game, it's clear that the resulting demo is different from, say, "-complevel 9 -record mydemo".

fabiangreffrath:
The `-longtics` parameter has merely been added to support playing back and recording demos in a format introduced by an EXE hack some decades ago. As @kraflab already stated, there is no other valid Doom demo format with long tics available. Also, just bumping the version number and calling it "MBF with longtics" is probably a bit short-sighted. For example, hasn't there been a source port called Fusion derived from MBF which claimed this number for itself? And what version number does Eternity Engine currently use?

ceski-1:
I can't think of a solution for now. Maybe I'll try again if I can come up with something.

PR #222, opened 2021-02-26
Fix automap markers shifting in widescreen

### LINKED ISSUE #90: Automap markers move incorrectly in 16:9 ratio
**Expected**

Automap markers move with the rest of the automap in follow mode in every direction when the player moves - regardless of the aspect ratio.

**Actual**

The map markers appear to "hover" or "slide" over the map in 16:9 ratio either towards "East" or "West" depending on which direction the player is headed. It resembles the parallax scrolling effect when moving. Especially noticeable on larger maps. It makes the feature almost unusable in 16:9 aspect ratio.

When the aspect ratio is set (or detected) to be 4:3 the markers appear to work correctly.
When only moving "North" or "South" (according to the automap) then the markers appear to move correctly in both aspect ratios.

**How to reproduce it**

1. `glboom-plus -iwad DOOM2.WAD -warp 12 -nomonsters`
2. Set aspect ratio to 16:9
3. Press Tab
4. Press M
5. Move right and left (on the map)
6. Notice the automap mark "hovering" right and left faster than the rest of the map
7. Move up and down (on the map)
8. Notice no "hovering"
9. Set aspect ratio to 4:3
10. Move right and left (on the map)
11. Notice no "hovering"

Please let me know if there's anything I should test or if I'm doing something wrong here.

jackrjli:
Fixes #90

### DIFF
--- a/prboom2/src/am_map.c
+++ b/prboom2/src/am_map.c
@@ -2141,13 +2141,13 @@ static void AM_drawMarks(void)
             {
               V_DrawNamePatchPrecise(
                 (float)p.fx * 320.0f / SCREENWIDTH, (float)p.fy * 200.0f / SCREENHEIGHT,
-                FB, namebuf, CR_DEFAULT, VPT_ALIGN_WIDE | VPT_STRETCH);
+                FB, namebuf, CR_DEFAULT, VPT_ALIGN_MAX | VPT_STRETCH);
             }
             else
             {
               V_DrawNamePatch(
                 p.x * 320 / SCREENWIDTH, p.y * 200 / SCREENHEIGHT,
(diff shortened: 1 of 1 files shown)


fabiangreffrath:
Hm, sorry, but the markers are still moving if I play at 1920x1080.

jackrjli:
Hmm, I can't reproduce the issue with this patch applied. I know that when using the software renderer the markers jitter around as you move the map, but they remain centered around the same point. (When using OpenGL am_frame.precise is set to true and the markers move more smoothly, and if you use the "icons" thing appearance setting, it uses a different function entirely for drawing the markers.)

However, I did just notice a different issue this created, which is that if you set the aspect ratio setting to 5:4, the positions of the markers get changed compared to all the other aspect ratio settings.

PR #230, opened 2021-03-02
show the current and next level when typing the IDCLEV cheat

### LINKED ISSUE #221: [REQUEST] Show map number on Automap
As some custom WADs do not have the map number in the custom map names, I would very much appreciate if the map number (e.g. MAP01 or E1M1) would show up in the Automap above the map name.

Thank you for reading.

fabiangreffrath:
Fixes #221

### DIFF
--- a/prboom2/src/m_cheat.c
+++ b/prboom2/src/m_cheat.c
@@ -46,6 +46,7 @@
 #include "d_deh.h"  // Ty 03/27/98 - externalized strings
 /* cph 2006/07/23 - needs direct access to thinkercap */
 #include "p_tick.h"
+#include "e6y.h" // G_GotoNextLevel()
 #include "w_wad.h" // W_GetLumpInfoByNum()
 
 #define plyr (players+consoleplayer)     /* the console player */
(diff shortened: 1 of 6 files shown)


fabiangreffrath:
Si, what do you say @snoggie1487, @runlow, @JadingTsunami ?

JadingTsunami:
Looks good! I might only recommend to leave a comment on the function behavior or change the prototype name from "Goto" to something more descriptive now that the behavior is changed.

Maybe "G_StoreOrGotoNextLevel()" or something?

fabiangreffrath:
> Looks good! I might only recommend to leave a comment on the function behavior or change the prototype name from "Goto" to something more descriptive now that the behavior is changed.

I don't think that I'll rename the function, its main purpose is still unchanged. But I agree that the new "reporting" behavior should be better commented.

prboom2/src/m_cheat.c
```diff
@@ -133,6 +136,7 @@ cheatseq_t cheat[] = {
   CHEAT("idbeholdl",  "Lite-Amp Goggles", not_dm, cheat_pw, pw_infrared),
   CHEAT("idbehold",   "BEHOLD menu",      not_dm, cheat_behold, 0),
   CHEAT("idclev",     "Level Warp",       cht_never | not_menu, cheat_clev, -2),
+  CHEAT("idclev",     NULL,               cht_never | not_menu, cheat_clev0, 0),
```
fabiangreffrath, line 139: I am not sure if this should read "Level Warp" as well? Else, if the cheat sequence is replaced by DEHACKED, it will print the level information when IDCLEV is typed but won't warp to the entered level and if the replacement cheat sequence for IDCLEV is typed it will warp without showing this information at all.

runlow:
Seems to work great at f0a1be3a

PR #233, opened 2021-03-03
Fix hanging notes on exit when using Portmidi (fixes #232)

### LINKED ISSUE #232: PortMIDI backend won't stop notes after exiting game
The PortMIDI music backend is simple and good enough - it sends MIDI events to an output MIDI sink in the system.

There is one issue with that - it doesn't stop ("panic") MIDI channels after the program exits, which often leads to notes that don't end.

In my particular setup (w/ Manjaro Linux), I use ALSA for MIDI routing, and run a FluidSynth daemon with systemd and an user service. Then, using QjackCtl's graph, I connect the MIDI Through input to the FluidSynth output port. MIDI Through is in its essence a loopback, and thus what this does is it plugs all events that go to MIDI Through back and through FluidSynth. Yes, I have defined a QjackCtl patchbay configuration so it does this connection automatically for me now. Just some context, don't think any of these are the culprit.

An easy solution would be to send a MIDI panic on all channels when PrBoom+ exits whilst using the PortMIDI backend. Make sure to do this on all shutdowns, including I_Error ones! If anything, just check if a song was playing at this current state (if not, no need to send panics).

Gustavo6046:
When using the PortMidi player, e.g. in an ALSA MIDI setup, closing the program often [leaves some hanging notes](https://github.com/coelckers/prboom-plus/issues/232), since they aren't stopped by the game when it exits.

This hopefully fixes it by sending a MIDI ~~Stop All Sound~~ All Notes Off message to all channels in the shutdown callback, right before actually shutting PortMidi itself down, and by allowing the delay between that and the Portmidi shutdown to happen on non-Win32 platforms too.

Also, for the sake of linkage. Fixes #232 .

### DIFF
--- a/prboom2/src/MUSIC/portmidiplayer.c
+++ b/prboom2/src/MUSIC/portmidiplayer.c
@@ -185,7 +185,7 @@ static int pm_init (int samplerate)
 
 #endif
 
-static void pm_killnotes (unsigned long when = 0)
+static void pm_killnotes (unsigned long when)
 {
   /*
   workaround to hanging notes with portmidi; send "All Sound Off"
@@ -234,7 +234,7 @@ static void pm_shutdown (void)
     #endif
 
 	// stop all sound, in case of hanging notes
-    pm_killnotes();
+    pm_killnotes(0);
 
     Pm_Close (pm_stream);
     Pm_Terminate ();
--- a/prboom2/src/MUSIC/portmidiplayer.c
+++ b/prboom2/src/MUSIC/portmidiplayer.c
@@ -208,9 +208,7 @@ static void pm_shutdown (void)
     
     not a fix: calling Pm_Abort(); then midiStreamStop deadlocks instead of midiStreamClose. 
     */
-    #ifdef _WIN32
     Pt_Sleep (DRIVER_LATENCY * 2);
-    #endif
 
     Pm_Close (pm_stream);
     Pm_Terminate ();
--- a/prboom2/src/MUSIC/portmidiplayer.c
+++ b/prboom2/src/MUSIC/portmidiplayer.c
@@ -178,6 +178,8 @@ static int pm_init (int samplerate)
   return 1;
 }
 
+static void pm_stop (void);
+
 static void pm_shutdown (void)
 {
   if (pm_stream)
--- a/prboom2/src/MUSIC/portmidiplayer.c
+++ b/prboom2/src/MUSIC/portmidiplayer.c
@@ -390,7 +390,7 @@ static void writesysex (unsigned long when, int etype, unsigned char *data, int 
 static void pm_stop (void)
 {
   int i;
-  unsigned long when = Pt_Time ();
+  unsigned long when = 0;
   pm_playing = 0;
   
 
--- a/prboom2/src/MUSIC/portmidiplayer.c
+++ b/prboom2/src/MUSIC/portmidiplayer.c
@@ -178,6 +178,32 @@ static int pm_init (int samplerate)
   return 1;
 }
 
+#ifndef MIDI_EVENT_CHANNEL_MODE
+
+// channel mode events in MIDI, low nibble is channel value as usual
+#define MIDI_EVENT_CHANNEL_MODE 0xB0
+
+#endif
+
+static void pm_killnotes (unsigned long when = 0)
+{
+  /*
+  workaround to hanging notes with portmidi; send "All Sound Off"
+  events to all channels when the game exits
+
+  see: http://personal.kent.edu/~sbirch/Music_Production/MP-II/MIDI/midi_channel_mode_messages.htm
+
+  the event would be Bn 78 00, where 'n' is a MIDI channel; eg. the
+  MIDI message B1 78 00 stops all sound in channel 2 (of 1-16)
+  */
+
+  when = when | (Pt_Time() * !when); // set when to Pt_Time() if when is zero
+
+  for (int ch = 0; ch < 16; ch++) {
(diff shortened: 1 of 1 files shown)


prboom2/src/MUSIC/portmidiplayer.c
```diff
@@ -178,6 +178,32 @@ static int pm_init (int samplerate)
...
 }
 
+#ifndef MIDI_EVENT_CHANNEL_MODE
+
+// channel mode events in MIDI, low nibble is channel value as usual
+#define MIDI_EVENT_CHANNEL_MODE 0xB0
+
+#endif
+
+static void pm_killnotes (unsigned long when = 0)
```
alexey-lysiuk, line 188: C language doesn't support default values, it's a C++ thing.

Gustavo6046: Yeah, I noticed that with the CI failure. I was just pushing commits to fix that :p

Gustavo6046:
Build seems to be working now. I'll do on my end too. If all goes well I should be able to test in a moment.

Gustavo6046:
Unfortunately it doesn't stop hanging notes from occurring. I am not sure why yet. I will investigate.

Gustavo6046:
`pm_killnotes` _is_ called, according to gdb. So the issue is probably with the event sent, or how it is sent, or maybe it's the fact that `Pm_Close` is called right after `Pm_Write` (`writeevent`)? Maybe it doesn't _actually_ send the event because of that? I'm not sure. Is there a `Pm_Flush` type function I could call to work around this, at least for now?

prboom2/src/MUSIC/portmidiplayer.c
```diff
@@ -207,6 +232,9 @@ static void pm_shutdown (void)
     Pt_Sleep (DRIVER_LATENCY * 2);
     #endif
 
+	// stop all sound, in case of hanging notes
+    pm_killnotes(0);
```
fabiangreffrath, line 236: Couldn't you just call `pm_stop()` here?

Gustavo6046: Good point, I'll do that.

Gustavo6046: It also seems insufficient - hanging notes still occur. I'll investigate further.

Gustavo6046:
Thanks Fabian, pm_stop does seem to be precisely what pm_killnotes aspired to be :p

Unfortunately, it still doesn't resolve the issue.

Perhaps it isn't a prboom-plus issue - `aplaymidi` also doesn't stop notes if it is interrupted prematurely with Ctrl+C (aka `sigint`), although that might be by design since it's meant to play MIDI events directly, and probably not supposed to be a MIDI player of sorts.

On the other hand, Drumstick doesn't seem to be affected by this issue. It's a simple MIDI player, so I suppose it is expected to stop ongoing notes if it closes before the end of a song.

Perhaps a better "fix" would be to just add an ALSA MIDI backend for Linux builds? Hmm.

Gustavo6046:
You know what, scratch Portmidi. Feel free to close this PR if you feel like it is going nowhere. My efforts might be better invested into implementing an ALSA MIDI backend. If all goes well I might even add a JACK one in the near future, or even a TCP one. I dunno.

fabiangreffrath:
Hm, I am surprised the latest commit doesn't fix the issue for you. Would you mind having a look at the shutdown sequence in Drumstick, i.e. which portmidi functions are called there?

Gustavo6046:
I'm not sure if Drumstick uses Portmidi. If it doesn't, there's our answer.

I'm currently working on an ALSA interface, which is basically a weird port of portmidiplayer.c to the [ALSA Sequencer API](https://www.alsa-project.org/alsa-doc/alsa-lib/seq.html). If all goes well, we might have this in the preference list above Portmidi! (CMakeLists.txt will look for ALSA, so don't worry about portability, as it'll be skipped for non-Linux users anyway.)

fabiangreffrath:
I am still not giving up on this one. It might as well be a timing problem: The "All Notes Off" command is sent to a pm "device" that is shut down in the very same moment.

Could you try setting `when` to `0` in `pm_stop()` (which means *now* if I am not mistaken), please?

Gustavo6046:
Setting `when` to `0` inside `pm_stop()` sadly does not fix the issue.

fabiangreffrath:
What if you remove the condition around that `Pt_Sleep()` call a few lines farther?

Gustavo6046:
> What if you remove the condition around that `Pt_Sleep()` call a few lines farther?

Well, that seems to fix it!

That does make the semantics of the comment preceding it a tad silly. :P

Gustavo6046:
Also, for the sake of linkage. Fixes #232 .

prboom2/src/MUSIC/portmidiplayer.c
```diff
@@ -385,7 +388,7 @@ static void writesysex (unsigned long when, int etype, unsigned char *data, int
 static void pm_stop (void)
 {
   int i;
-  unsigned long when = Pt_Time ();
+  unsigned long when = 0;
```
fabiangreffrath, line 391: I guess it's safe to revert this change back to the original code.

fabiangreffrath (changes requested):

prboom2/src/MUSIC/portmidiplayer.c
Gustavo6046: I think so too.

fabiangreffrath:
Thank you!

Gustavo6046:
Why of course :)

PR #234, opened 2021-03-03
Add ALSA sequencer API backend for ALSA MIDI support

Gustavo6046:
This is an experimental implementation of the MIDI player that utilizes ALSA's sequencer interface in order to create a MIDI source ALSA port. This is absolutely untested, as of opening this pull request.

It's meant to replace the Portmidi backend, which is flawed on Linux systems. After all, Portmidi under Linux tends to use ALSA anyway. The only exception is if ALSA is not available, but that should generally not be the case, since ALSA is mature and has been around for a good while; OSS is legacy, outdated, and everyone who uses Linux uses ALSA for their MIDI affairs.

For more pretext on why I am doing this, see issue #232 and pull request #233.

### DIFF
--- a/prboom2/src/MUSIC/alsaplayer.c
+++ b/prboom2/src/MUSIC/alsaplayer.c
@@ -318,7 +318,7 @@ int alsa_midi_default_dest (void)
         continue;
       }
 
-      if (loopback_cli == client_num)
+      if (loopback_cl == client_num)
       {
         // save as midi through port
 
--- a/prboom2/src/MUSIC/alsaplayer.c
+++ b/prboom2/src/MUSIC/alsaplayer.c
@@ -199,6 +199,8 @@ void alsaplay_refresh_outputs(void) {
       alsaplayer_outputs[out_ind].client = client_num;
       alsaplayer_outputs[out_ind].port = port_num;
 
+      lprintf(LO_INFO, "alsaplay_refresh_outputs: output #%d: (%d:%d) %s\n", out_ind, client_num, port_num, client_name);
+
       // client name only up to 100 chars, so it always fits within a 120 byte buffer
       sprintf(alsaplayer_outputs[out_ind].name, "%.*s (%d:%d)", 100, snd_seq_client_info_get_name(cinfo), client_num, port_num);
     }
@@ -635,6 +637,8 @@ static int alsa_init (int samplerate)
   }
 
   lprintf (LO_INFO, "alsaplayer: Successfully opened port: %d\n", out_port);
+
+  alsaplay_refresh_outputs(); // make output list and print it out
 
   return 1;
 }
--- a/prboom2/src/MUSIC/alsaplayer.c
+++ b/prboom2/src/MUSIC/alsaplayer.c
@@ -182,6 +182,11 @@ void alsaplay_refresh_outputs(void) {
 
       // check if port is valid midi output
 
+      if (!(snd_seq_port_info_get_type(pinfo) & SND_SEQ_PORT_TYPE_MIDI_GENERIC))
+      {
+        continue;
+      }
+
       if (snd_seq_port_info_get_capability(pinfo) & OUT_CAPS_DESIRED != OUT_CAPS_DESIRED)
       {
         continue;
@@ -315,6 +320,11 @@ int alsa_midi_default_dest (void)
       int port_num = snd_seq_port_info_get_port(pinfo);
 
       // check if port is valid midi output
+
+      if (!(snd_seq_port_info_get_type(pinfo) & SND_SEQ_PORT_TYPE_MIDI_GENERIC))
+      {
+        continue;
+      }
 
       if (snd_seq_port_info_get_capability(pinfo) & OUT_CAPS_DESIRED != OUT_CAPS_DESIRED)
       {
--- a/prboom2/src/MUSIC/alsaplayer.c
+++ b/prboom2/src/MUSIC/alsaplayer.c
@@ -252,7 +252,7 @@ int alsa_midi_default_dest (void)
   alsaplay_clear_outputs();
 
   int loopback_cl = -1, loopback_prt = 0;
-  char *loopback_name;
+  const char *loopback_name;
 
   // clear client info
   snd_seq_client_info_set_client(cinfo, -1);
--- a/prboom2/src/MUSIC/alsaplayer.c
+++ b/prboom2/src/MUSIC/alsaplayer.c
@@ -420,6 +420,7 @@ int alsa_midi_set_dest (int client, int port)
   CHK_LPRINT_ERR_RET(snd_seq_connect_to(seq_handle, out_port, client, port), -3,
     LO_WARN, "alsa_midi_set_dest: error connecting to (%d:%d): %s", last_client, last_port);
 
+  alsa_first_connected = 1;
   return 0;
 }
 
@@ -577,6 +578,28 @@ static void alsa_midi_all_notes_off (void)
   }
 }
 
+static int alsa_first_connected = 0;
+static void alsa_midi_init_connect_default_port (void)
+{
+  // load MIDI device specified in config
+
+  if (snd_mididev && strlen(snd_mididev))
+  {
+    snd_seq_addr_t seqaddr;
+
+    CHK_LPRINT_ERR(snd_seq_parse_address(seq_handle, &seqaddr, snd_mididev),
+      LO_WARN, "alsa_init: Error connecting to configured MIDI output port \"%s\": %s", snd_mididev)
+
+    return alsa_midi_set_dest(seqaddr.client, seqaddr.port) == 0;
+  }
+
+  else
+  {
+    // connect to default
+    return alsa_midi_default_dest();
+  }
+}
+
 ////////////////////
 
 // alsa player callbacks
@@ -598,24 +621,6 @@ static int alsa_init (int samplerate)
   }
 
   lprintf (LO_INFO, "alsaplayer: Successfully opened port: %d\n", out_port);
-
-  // load MIDI device specified in config
-
-  if (snd_mididev && strlen(snd_mididev))
-  {
-    snd_seq_addr_t seqaddr;
-
-    CHK_LPRINT_ERR(snd_seq_parse_address(seq_handle, &seqaddr, snd_mididev),
-      LO_WARN, "alsa_init: Error connecting to configured MIDI output port \"%s\": %s", snd_mididev)
-
-    return alsa_midi_set_dest(seqaddr.client, seqaddr.port) == 0;
-  }
-
-  else
-  {
-    // connect to default
-    return alsa_midi_default_dest();
-  }
 
   return 1;
 }
@@ -766,6 +771,13 @@ static void alsa_resume (void)
 static void alsa_play (const void *handle, int looping)
 {
   // reinit queue
+
+  if (!alsa_first_connected)
+  {
+    // connect to default port if haven't connected at least once yet
+    alsa_midi_init_connect_default_port();
+  }
+
   if (out_queue) {
     snd_seq_free_queue(seq_handle, out_queue);
     snd_seq_queue_status_free(queue_status);
--- a/prboom2/src/MUSIC/alsaplayer.c
+++ b/prboom2/src/MUSIC/alsaplayer.c
@@ -130,9 +130,12 @@ void alsaplay_clear_outputs(void) {
   alsaplayer_num_outs = 0;
 }
 
+static snd_seq_client_info_t *cinfo;
+static snd_seq_port_info_t   *pinfo;
+
 void alsaplay_refresh_outputs(void) {
-  static snd_seq_client_info_t cinfo;
-  static snd_seq_port_info_t   pinfo;
+  snd_seq_client_info_malloc(&cinfo);
+  snd_seq_port_info_malloc  (&pinfo);
 
   // port type and capabilities required from valid MIDI output
   const int OUT_CAPS_DESIRED = (SND_SEQ_PORT_CAP_SUBS_WRITE);
@@ -192,6 +195,9 @@ void alsaplay_refresh_outputs(void) {
       sprintf(&alsaplayer_outputs[out_ind].name, "%.*s (%d:%d)", 100, snd_seq_client_info_get_name(&cinfo), client_num, port_num);
     }
   }
+
+  snd_seq_client_info_free(cinfo);
+  snd_seq_port_info_free  (pinfo);
 }
 
 int alsaplay_connect_output(int which) {
--- a/prboom2/src/MUSIC/alsaplayer.c
+++ b/prboom2/src/MUSIC/alsaplayer.c
@@ -802,6 +802,38 @@ static void alsa_play (const void *handle, int looping)
   
   snd_seq_queue_status_malloc(&queue_status);
 
+  // set queue resolution
+  snd_seq_queue_timer_t *timer;
+
+  snd_seq_queue_timer_malloc(&timer);
+
+  int status = snd_seq_get_queue_timer(seq_handle, out_queue, timer);
+
+  if (status < 0)
+  {
+    lprintf(LO_WARN, "alsa_play: error getting sched queue timer: %s\n", snd_strerror(status));
+
+    snd_seq_queue_timer_free(timer);
+    goto finish;
+  }
+
+  snd_seq_queue_timer_set_resolution(timer, 1000000 / 32); // 1000000 ns = 1 ms, so this is 1/32 ms
+
+  status = snd_seq_set_queue_timer(seq_handle, out_queue, timer);
+
+  if (status < 0)
+  {
+    lprintf(LO_WARN, "alsa_play: error setting sched queue timer with new resolution: %s\n", snd_strerror(status));
+
+    snd_seq_queue_timer_free(timer);
+    goto finish;
+  }
+
+  lprintf(LO_INFO, "alsa_play: success\n");
+
+  snd_seq_queue_timer_free(timer);
+
+finish:
   // initialize state stuff
   eventpos = 0;
   alsa_looping = looping;
--- a/prboom2/src/MUSIC/alsaplayer.c
+++ b/prboom2/src/MUSIC/alsaplayer.c
@@ -597,6 +597,7 @@ static void alsa_resume (void)
 
   snd_seq_continue_queue(seq_handle, out_queue, 0);
 }
+
 static void alsa_play (const void *handle, int looping)
 {
   // reinit queue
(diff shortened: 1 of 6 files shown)


fabiangreffrath:
> everyone who uses Linux uses ALSA for their MIDI affairs.

*cough* pulseaudio 😉

Gustavo6046:
The ALSA backend finally seems to be in a.. working state. It's pretty experimental, and I still have to see if there are other places where flushing the queue right away would also be desirable.

But the nasty parts are already over! Goodbye Portmidi, hello ALSA! At last we have a relatively maintainable ALSA backend that works to a degree.

@fabiangreffrath apologies if I got anything wrong. It was my first time working with MIDI directly, as you can probably tell from this PR's commit log. It took quite a bit of fiddling to have it work well enough. I really hope it's up to standards, too!

Gustavo6046:
And now, thanks to that one last commit, the ALSA backend is not prone to the hanging notes issue that prompted these two pull requests to begin with. :D

Gustavo6046:
Note: this backend does not set up any subscriptions automatically. At the moment, a patchbay (like QjackCtl's graph) is needed to connect this to an ALSA MIDI destination port, e.g. FluidSynth.

Gustavo6046:
> > everyone who uses Linux uses ALSA for their MIDI affairs.
> 
> _cough_ pulseaudio wink

PulseAudio doesn't route MIDI, only audio. :P

prboom2/src/SDL/i_sound.c
```diff
@@ -1504,6 +1508,7 @@ static int Exp_RegisterSongEx (const void *data, size_t len, int try_mus2mid)
           }
           else
             lprintf (LO_INFO, "Exp_RegisterSongEx: Music player %s on preferred list but it failed to init\n", music_players[i]-> name ());
+          break;
```
fabiangreffrath, line 1511: Why do you introduce this `break` here?

prboom2/src/m_cheat.c
```diff
@@ -194,12 +195,38 @@ cheatseq_t cheat[] = {
   // Enable/disable shorttics in-game
   CHEAT("tntshort",   NULL,               cht_never, cheat_shorttics, 0),
 
+  // subscribe to dest alsa port manually
```
fabiangreffrath, line 198: Sorry, but this is unacceptable. A music backend that requires typing a cryptic cheat to work at all will go down in history as the one that nobody used, even on Linux, even with ALSA midi setup.

fabiangreffrath (changes requested):

prboom2/src/SDL/i_sound.c
Gustavo6046: Because the following iterations of this loop are wasted. It doesn't matter much, but it was a bit annoying when debugging with gdb. :p

prboom2/src/m_cheat.c
Gustavo6046: It's for debugging purposes, it's not meant to be used by the end user. I can remove it if you prefer. I still have to find a way to list possible destination MIDI ports in ALSA, though.

Gustavo6046:
Those are very fair and valid points! I'll take a look at them tomorrow.

Gustavo6046:
Today I can look into finding MIDI destination ports from the code itself.

fabiangreffrath:
What advantages may we expect compared to the portmidi backend?

alexey-lysiuk (commented):
Please add installation of the corresponding dependency to [Linux continuous integration](https://github.com/coelckers/prboom-plus/blob/ce65528a95393d0391058c1ba98ca33c12d6e86c/.github/workflows/continuous_integration.yml#L69-L70). This helps to validate that new code compiles.

Gustavo6046:
> What advantages may we expect compared to the portmidi backend?

Portmidi is a layer of abstraction. There are inherent benefits with providing direct interfaces as well. For instance, the code has more control and flexibility over _how_ we play and route our MIDI events. It's also very useful when troubleshooting music in Linux, say, if Portmidi chooses a different MIDI backend for some reason and bugs out because of that.

Gustavo6046:
> Please add installation of the corresponding dependency to [Linux continuous integration](https://github.com/coelckers/prboom-plus/blob/ce65528a95393d0391058c1ba98ca33c12d6e86c/.github/workflows/continuous_integration.yml#L69-L70). This helps to validate that new code compiles.

Indeed, it's important to have this in CI as well. How could I forget this! I even set up the ALSA library in CMakeLists.txt and even specified the library in the INSTALL file!

Also, in case you were wondering, I was testing it locally.

Gustavo6046:
I'll turn this PR into a draft, for now, until I get the port listing logic done.

@fabiangreffrath how should I take the port configuration from the user? Is there some configuration I can read values from?

Gustavo6046:
I think I found the functions I need, from the [Sequencer Client](https://www.alsa-project.org/alsa-doc/alsa-lib/group___seq_client.html) and [Sequencer Port](https://www.alsa-project.org/alsa-doc/alsa-lib/group___seq_port.html) interfaces.

Gustavo6046:
@fabiangreffrath Now there is code to query all clients and their output ports. Now how do I integrate this with the game so the user can select a port? Maybe do it in the options menu?

Maybe a simple cheat interface could suffice for now. `ALSNUM` would print the number of outputs available, `ALSGETxx` would print "name (id:port)" of listed output #xx and `ALSUSExx` would use it. "xx" would start at 1 rather than 0. This would be just a temporary makeshift thing, until we can decide on something better. What do you think?

fabiangreffrath:
Nope. Cheat codes are not a valid configuration interface for a music backend. Please keep these out of the game. 

On an average system, how many different outputs and ports do we expect? If there's only one, use this. If in doubt, use the first. Or add a config key that holds a simple `xxx:y` string, defaulting to `000:0`.

Gustavo6046:
> Nope. Cheat codes are not a valid configuration interface for a music backend. Please keep these out of the game.

Oh, alright. I will then remove them once we consolidate a better interface. I'll only keep them around _for now_ for testing purposes, if that is okay with you?

> On an average system, how many different outputs and ports do we expect? If there's only one, use this. If in doubt, use the first. Or add a config key that holds a simple `xxx:y` string, defaulting to `000:0`.

It really depends. Usually we expect a MIDI Through output and a synthesizer one (e.g. a soft synth, like a Fluidsynth or Timidity daemon, or a hardware MIDI output). Software synthesizers can actually open multiple ports. In particular Timidity does this, not sure why.

But yes, I can totally write code that defaults to the lowest client/port id combination that isn't MIDI Through. (Maybe a case-insensitive strcmp for the first characters of the client name being "midi through" could be enough to rule it out here?) And if none are found, I guess warn in stderr or something, or if preferred swap to another music player altogether.

And, what do you mean config key? My apologies, I haven't quite familiarized myself with the configuration/menu side of the PrBoom+ codebase.

Oh, and also, can we buffer some per-channel MIDI status (ie program changes, etc)? This way, we can send them again when switching to a different output, so we can preserve the instruments and such.

Gustavo6046:
I was thinking, if the codes work, at least we can know that the new output listing functionality in `alsaplayer.c` works, and that we may confidently use it when the time comes to write an actual menu for selecting a MIDI output or something.

And yeah, you may rest assured that the codes **will** be disabled before I set this PR as ready for review again (aka unset draft).

Gustavo6046:
Wow, this actually worked even better than I was expecting! ALSNUM says there are 2 outputs available. ALSGET01 states that #1 is MIDI Through. ALSGET02 correctly states that #02 is Fluidsynth. ALSUSE01 makes no sound, but ALSUSE02 does pipe the music through Fluidsynth accordingly. Nice!

On the other hand, the underlying function that connects to an output does not reset MIDI controllers like instruments, like I said above, resulting on forcing every channel into the Grand Piano. Hey, D_RUNNIN sounds pretty good on the piano keyboard if you ask me! :p

fabiangreffrath:
> Oh, alright. I will then remove them once we consolidate a better interface. I'll only keep them around _for now_ for testing purposes, if that is fine with you?

Well, let's say I will definitely not merge this PR with the cheat codes in place. As we all know, nothing lasts longer than a "temporary solution" :wink: 

> But yes, I can totally write code that defaults to the lowest client/port id combination that isn't MIDI Through. (Maybe a case-insensitive strcmp for the first characters of the client name being "midi through" could be enough to rule it out here?) And if none are found, I guess warn in stderr or something, or if preferred swap to another music player altogether.

The most important part is that this backend works out of the box - with as little user configuration required as possible. We need reasonable defaults!

> And, what do you mean config key? My apologies, I haven't quite familiarized myself with the configuration/menu side of the PrBoom+ codebase.

Erm, I mean one of the key/value pairs that make up `prboom-plus.cfg`. There are some exclusive to the fluidsynth backend, so I see no reason why we shouldn't configure the ALSA backend this way, too:

https://github.com/coelckers/prboom-plus/blob/ce65528a95393d0391058c1ba98ca33c12d6e86c/prboom2/src/m_misc.c#L441

> Oh, and also, can we buffer some per-channel MIDI status (ie program changes, etc)? This way, we can send them again when switching to a different output, so we can preserve the instruments and such.

I am not sure what you are asking here, honestly. But I'd appreciate if this code could remain as close as possible to what it is supposed to be: a MIDI backend for prboom-plus - a means to listen to music while playing a game. Let's please resists to turn this into a playground for MIDI enthusiasts. And I'd also appreciate if this could touch unrelated code as little as possible.

prboom2/src/SDL/i_sound.c
```diff
@@ -1305,14 +1308,15 @@ char music_player_order[NUM_MUS_PLAYERS][200] =
   PLAYER_DUMB,
   PLAYER_FLUIDSYNTH,
   PLAYER_OPL2,
+  PLAYER_ALSA,
```
fabiangreffrath, line 1311: Please stand in line. The portmidi backend is here for decades, it is known working and does so out-of-the-box. Also is the new backend, with no working default configuration yet and currently tested by only one user - its developer. Please add it last, thanks!

fabiangreffrath (changes requested):
Please sort "alsa" after "portmidi" throughout the code, thanks!

prboom2/src/SDL/i_sound.c
Gustavo6046: Indeed. This is an experimental backend, it shouldn't be preferred over a true and tested one like Portmidi. I should have put more thought into the ordering. I'll swap it into place.

Gustavo6046:
> > Oh, alright. I will then remove them once we consolidate a better interface. I'll only keep them around _for now_ for testing purposes, if that is fine with you?
> 
> Well, let's say I will definitely not merge this PR with the cheat codes in place. As we all know, nothing lasts longer than a "temporary solution" wink

Well, yes, I expect that and am aware of that! Of course I'm going to remove them once the ALSA backend  A) has reasonable defaults, and B) is configurable by the user.

> > But yes, I can totally write code that defaults to the lowest client/port id combination that isn't MIDI Through. (Maybe a case-insensitive strcmp for the first characters of the client name being "midi through" could be enough to rule it out here?) And if none are found, I guess warn in stderr or something, or if preferred swap to another music player altogether.
> 
> The most important part is that this backend works out of the box - with as little user configuration required as possible. We need reasonable defaults!

That is the goal here :)

> > And, what do you mean config key? My apologies, I haven't quite familiarized myself with the configuration/menu side of the PrBoom+ codebase.
> 
> Erm, I mean one of the key/value pairs that make up `prboom-plus.cfg`. There are some exclusive to the fluidsynth backend, so I see no reason why we shouldn't configure the ALSA backend this way, too:
> 
> https://github.com/coelckers/prboom-plus/blob/ce65528a95393d0391058c1ba98ca33c12d6e86c/prboom2/src/m_misc.c#L441

Ah, I see. Yes, I can do that. It'll be pretty easy. In fact, ALSA provides an utility function to parse a string denoting an output port. We can use that.

> > Oh, and also, can we buffer some per-channel MIDI status (ie program changes, etc)? This way, we can send them again when switching to a different output, so we can preserve the instruments and such.
> 
> I am not sure what you are asking here, honestly. But I'd appreciate if this code could remain as close as possible to what it is supposed to be: a MIDI backend for prboom-plus - a means to listen to music while playing a game. Let's please resists to turn this into a playground for MIDI enthusiasts. And I'd also appreciate if this could touch unrelated code as little as possible.

What I mean is that in-game music uses MIDI events to set the instruments at the beginning of the song (and, sometimes, in the middle), but this state isn't recorded at all - it's only recorded by the actual synthesizer at the other end of the MIDI connection (like Fluidsynth or Timidity), and if we're going to have the ability to switch to a different port at run-time, maybe we should store the channel state, since otherwise the notes will be sent on channels with the default instrument (in MIDI that is the Grand Piano). I'm sure other controller states would need to be saved as well, although I'm not entirely sure of which.

So, tl;dr switching ports will reset all instruments if we don't save them ourselves and resend them whenever we connect to a different port.

Yes, this does assume switching output ports at run-time. But one of the goals I initially had with this backend (and that in a way I've had moderate success so far, with ALSUSE), was to be able to switch to a different port _during_ the game, something not even the Portmidi backend could do.

What if one starts a software synthesizer, or even plugs in a hardware one, _after_ PrBoom+ starts? I mean, it's not all that important, and not a likely scenario either, but I don't like to cut corners. :)

Gustavo6046:
I added some logic to find a default port, as well as connecting to a port specified in configuration (if any). The latter is shared with Portmidi, because `snd_mididev` is not specialized in any MIDI interface. I suppose that is fine?

Gustavo6046:
Oops.

Gustavo6046:
CI is having issues with downloading files or something.

Gustavo6046:
@fabiangreffrath I managed to implement automatic default port selection, as well as reading it from snd_mididev if the latter is set (and non-empty). I also removed the cheats. Hopefully this means it does what you wanted it to :)

Also, CI is undergoing some strange errors in the Install Dependencies stage. What is that about?

Gustavo6046:
Ah, "mirror sync in progress?" the command said. And now Linux Clang worked, after the empty commit. Maybe another refreshing nudge will help... :)

Gustavo6046:
Keep it in mind that the only two CI builds we're interested in are Linux GCC and Linux Clang. ALSA is after all a Linux thing :D

alexey-lysiuk:
I prefer to have `m_cheat.c` changes in a separate PR. Apparently, finished ALSA support will be squashed into one commit.

Gustavo6046:
> I prefer to have `m_cheat.c` changes in a separate PR. Apparently, finished ALSA support will be squashed into one commit.

I thought I reverted those already. I'll fix that.

Gustavo6046:
There we go.

Gustavo6046:
I wish there were more people who could test this in their Linux systems apart from of course myself.

Gustavo6046:
I've come to the conclusion that what we have here works fine, but the stuttery issue arises most likely from the MUS format.

prboom2/src/SDL/i_sound.c
```diff
@@ -1282,6 +1283,7 @@ static const music_player_t *music_players[] =
   &db_player, // dumbplayer.h
   &fl_player, // flplayer.h
   &opl_synth_player, // oplplayer.h
+  &alsa_player, // alsaplayer.h
```
fabiangreffrath, line 1286: Is the order important here?

prboom2/src/MUSIC/alsaplayer.h
```diff
@@ -0,0 +1,61 @@
+/* Emacs style mode select   -*- C++ -*-
+ *-----------------------------------------------------------------------------
+ *
+ *
+ *  PrBoom: a Doom port merged with LxDoom and LSDLDoom
+ *  based on BOOM, a modified and improved DOOM engine
+ *
+ *  Copyright (C) 2011 by
```
fabiangreffrath, line 8: Please add your copyright here.

prboom2/src/MUSIC/alsaplayer.c
```diff
@@ -0,0 +1,1041 @@
+/* Emacs style mode select   -*- C++ -*-
+ *-----------------------------------------------------------------------------
+ *
+ *
+ *  PrBoom: a Doom port merged with LxDoom and LSDLDoom
+ *  based on BOOM, a modified and improved DOOM engine
+ *
+ *  Copyright (C) 2011 by
```
fabiangreffrath, line 8: Please add your copyright here.

prboom2/src/MUSIC/alsaplayer.c
```diff
@@ -0,0 +1,1041 @@
...
+  return data;
+}
+
+/*
+portmidi has no overall volume control.  we have two options:
+1. use a win32-specific hack (only if mus_extend_volume is set)
+2. monitor the controller volume events and tweak them to serve our purpose
+*/
+
+#ifdef _WIN32
```
fabiangreffrath, line 704: _WIN32?

fabiangreffrath (changes requested):
LGTM in general, thank you very much for your effort!

I cannot say much about the actual implementation, though, as I am not familiar with the ALSA API at all, but I see this is now cleanly integrated in to the surrounding code. Nice.


However, I also see some minor issues. There is some boilerplate code left from the portmidi backend (even error messages that start with "portmidi:"), there are CPP checks for `_WIN32` around, the copyright holder and year are not updated and the code emits new warnings from both GCC and CLANG. Please clean this up a bit.

Also, there have been 115 (!) commits to this branch in less than a week. I'd prefer if we could let this calm down a bit before merging so chances are high we don't need to update the code at that rate once it's part of the master branch. So, I am waiting for your "go!" once you feel confident this has reached its final state (no matter if it has, but only you can tell if this feels ready or not).

prboom2/src/SDL/i_sound.c
Gustavo6046: It isn't. I don't know how I missed that instance in particular, but it still seemed to work the way it was supposed to in the UI (Portmidi before ALSA).

prboom2/src/MUSIC/alsaplayer.h
Gustavo6046: The original code was the Portmidi player, which it seems was written by Nicholai Main. Is it fine if I keep his name in there?

prboom2/src/MUSIC/alsaplayer.c
Gustavo6046: See above.

Gustavo6046:
Thank you!

Well, the commit rate is mostly me breaking up certain changes into parts that are not very relevant with each other. This would help revert them individually if only a specific functionality was to be restored, among other somewhat useful little things. I do agree that it'd be nice to condense them into a single commit when merging :P

prboom2/src/MUSIC/alsaplayer.c
Gustavo6046: Oops! Removing.

Gustavo6046:
I have cleaned the code up a bit. I don't know if it's good enough for a merge yet, but it does seem to work.

It does produce sound, even if certain music seems a bit off for some reason. I don't know if it's the timer resolution or what. It just can't seem to align the notes properly. The Portmidi backend, on the other hand, doesn't have this issue, which is both strange and concerning.

But it does play the music! I can't screenshot the sound, sadly, but I can show that it works in the UI as expected.
ALSA selected in the PrBoom+ General Settings interface

fabiangreffrath:
Done, thanks!

Gustavo6046:
Of course!

I wasn't still very satisfied with how the music plays a bit stuttered, but I never found a way to troubleshoot that, and I did look into it. Apart from that, I think it was pretty much good to go :)

PR #237, opened 2021-03-08
allow IDCLEV cheat and -warp parameter to warp to any available map

### LINKED ISSUE #235: [REQUEST] Allow IDCLEV-ing to maps above 32 if they exist
See: sf2012_final.

fabiangreffrath:
Fixes #235

### DIFF
--- a/prboom2/src/m_cheat.c
+++ b/prboom2/src/m_cheat.c
@@ -423,19 +423,8 @@ static void cheat_clev(char buf[3])
   entry = G_LookupMapinfo(epsd, map);
   if (!entry)
   {
+	  char name[9];
 
-	  // Catch invalid maps.
-	  if (epsd < 1 || map < 1 ||   // Ohmygod - this is not going to work.
-		  //e6y: The fourth episode for pre-ultimate complevels is not allowed.
-		  (compatibility_level < ultdoom_compatibility && (epsd > 3)) ||
-		  (gamemode == retail && (epsd > 4 || map > 9)) ||
-		  (gamemode == registered && (epsd > 3 || map > 9)) ||
-		  (gamemode == shareware && (epsd > 1 || map > 9)) ||
-		  (gamemode == commercial && (epsd > 1 || map > 33)))  //jff no 33 and 34
-		  return;                                                  //8/14/98 allowed
-
-	  if (!bfgedition && map == 33)
-		  return;
 	  if (gamemission == pack_nerve && map > 9)
 		  return;
 
@@ -444,6 +433,15 @@ static void cheat_clev(char buf[3])
 	  {
 		  epsd = 1;
 	  }
+
+	  if (gamemode == commercial)
+		  sprintf(name, "map%02d", map);
+	  else
+		  sprintf(name, "E%dM%d", epsd, map);
+
+	  // Catch invalid maps.
+	  if (W_CheckNumForName(name) == -1)
+		  return;
   }
   // So be it.
 
--- a/prboom2/src/m_cheat.c
+++ b/prboom2/src/m_cheat.c
@@ -386,19 +386,11 @@ struct MapEntry* G_LookupMapinfo(int gameepisode, int gamemap);
 // 'clev' change-level cheat
 static void cheat_clev0()
 {
-  char next[9];
+  char *next;
   int epsd, map;
 
   G_GotoNextLevel(&epsd, &map);
-
-  if (gamemode == commercial)
-  {
-    snprintf(next, 9, "MAP%02d", map);
-  }
-  else
-  {
-    snprintf(next, 9, "E%dM%d", epsd, map);
-  }
+  next = MAPNAME(epsd, map);
 
   doom_printf("Current: %s, Next: %s",  W_GetLumpInfoByNum(maplumpnum)->name, next);
 }
@@ -423,7 +415,7 @@ static void cheat_clev(char buf[3])
   entry = G_LookupMapinfo(epsd, map);
   if (!entry)
   {
-	  char name[9];
+	  char *name;
 
 	  if (gamemission == pack_nerve && map > 9)
 		  return;
@@ -434,10 +426,7 @@ static void cheat_clev(char buf[3])
 		  epsd = 1;
 	  }
 
-	  if (gamemode == commercial)
-		  sprintf(name, "map%02d", map);
-	  else
-		  sprintf(name, "E%dM%d", epsd, map);
+	  name = MAPNAME(epsd, map);
 
 	  // Catch invalid maps.
 	  if (W_CheckNumForName(name) == -1)
--- a/prboom2/src/s_sound.c
+++ b/prboom2/src/s_sound.c
@@ -233,6 +233,14 @@ void S_Start(void)
   memset(&musinfo, 0, sizeof(musinfo));
   musinfo.items[0] = -1;
 
+  // prevent music number under- and overflows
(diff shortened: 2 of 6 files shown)


prboom2/src/s_sound.c
```diff
@@ -233,6 +233,14 @@ void S_Start(void)
   memset(&musinfo, 0, sizeof(musinfo));
   musinfo.items[0] = -1;
 
+  // prevent music number under- and overflows
+  if (mnum <= mus_None || (gamemode == commercial && mnum <  mus_runnin) ||
+      mnum >= NUMMUSIC || (gamemode != commercial && mnum >= mus_runnin))
+  {
+    S_StopMusic();
```
kraflab, line 240: Instead of no music, should we have a default? Like taking the first map's midi for instance.

prboom2/src/m_cheat.c
```diff
@@ -444,6 +421,10 @@ static void cheat_clev(char buf[3])
 	  {
 		  epsd = 1;
 	  }
+
+	  // Catch invalid maps.
+	  if (W_CheckNumForName(MAPNAME(epsd, map)) == -1)
+		  return;
```
kraflab, line 427: How about a message to doom_printf like "idclev target 36 not found"

prboom2/src/s_sound.c
fabiangreffrath: In Crispy, I have some code to wrap the music number around back into the valid range if it either under- or overflows. But that's not canonical. Crashing would be canonical, but not desired, and silence somehow comes closest. 
But, honestly, I don't really like it either.

prboom2/src/m_cheat.c
fabiangreffrath: Yes, nice idea!

prboom2/src/s_sound.c
fabiangreffrath: Wait a minute. It seems like it's entirely impossible to reach a music overflow in Doom 1 (at least as long as we don't explicitly support two-digit maps - or only support them "by accident" like e.g. E1M10), because for episodes >= 4 all music will get mapped into episode 4's `spmus[]` music number array. So, we are left with Doom 2's music and for this I could wrap music numbers around back into the valid range - which doesn't make a difference for any map supported so far and prevents crashes for all the maps beyond that range.

fabiangreffrath: Something like this?
```patch
--- a/prboom2/src/s_sound.c
+++ b/prboom2/src/s_sound.c
@@ -181,6 +181,15 @@ void S_Stop(void)
 // Kills playing sounds at start of level,
 //  determines music if any, changes music.
 //
+
+static inline int WRAP(int i, int w)
+{
+  while (i < 0)
+    i += w;
+
+  return i % w;
+}
+
 void S_Start(void)
 {
   int mnum;
@@ -208,7 +217,7 @@ void S_Start(void)
     mnum = idmusnum; //jff 3/17/98 reload IDMUS music if not -1
   else
     if (gamemode == commercial)
-      mnum = mus_runnin + gamemap - 1;
+      mnum = mus_runnin + WRAP(gamemap - 1, NUMMUSIC - mus_runnin);
     else
       {
         static const int spmus[] =     // Song - Who? - Where?
@@ -225,22 +234,14 @@ void S_Start(void)
         };

         if (gameepisode < 4)
-          mnum = mus_e1m1 + (gameepisode-1)*9 + gamemap-1;
+          mnum = mus_e1m1 + WRAP((gameepisode-1)*9 + gamemap-1, mus_runnin - mus_e1m1);
         else
-          mnum = spmus[gamemap-1];
+          mnum = spmus[WRAP(gamemap-1, 9)];
       }

   memset(&musinfo, 0, sizeof(musinfo));
 ```

fabiangreffrath: Doom 1 isn't as safe as I thought. Forgot the ExM0 and E0My cases.

fabiangreffrath:
Last missing step is the map title widget. Episode being at least `1` is made sure in `G_InitNew()` and and `gamemap > 0` is made sure in `HU_Start()`, so I only need to check the upper bounds which is already being done for the Doom 2 variants, but nor for Doom 1.

fabiangreffrath:
So, what doesn't work:
* E0My: For some reason, the episode value is always set to at least 1 in `G_InitNew()` [1]. This is applied even outside the scope of UMAPINFO, so I guess there is a good reason for doing this.
* `-warp x 0`: Map  being passed as `0` to the `-warp` parameter has a special meaning [2], like "warp to the first new map in the PWAD" that I didn't want to change.

What should work:
* E1M10 (XBox IWAD): This works using `-warp 1 10` but not using IDCLEV, because the latter cannot handle two-digit map numbers.
* ExM0 (neis.wad), MAP00: This works using IDCLEVx0, but not using the `-warp` parameter, because map being `0` triggers a special meaning for the latter (see above).
* ExMy for 1 <= x <= 9 and 1 <= y <= 9 for both the `-warp` parameter and IDCLEV
* MAPxx for 01 <= xx <= 99 for both the `-warp` parameter and IDCLEV, but will probably crash if MAPxx+1 does not exists.

So, with "works" I mean that the map loads, that *some* music plays and that neither the map title widget nor the intermission screen crash. It does not mean "fully supported" in any way, but should be enough to check out that "secret map" in the PWAD. For arbitrary map names and transitions we will have to refer people to use UMAPINFO instead.


[1] https://github.com/coelckers/prboom-plus/blob/e9543d450f8dd0bddac98e0fcb41a80126275542/prboom2/src/g_game.c#L2889
[2] https://github.com/coelckers/prboom-plus/blob/e9543d450f8dd0bddac98e0fcb41a80126275542/prboom2/src/d_main.c#L2029

tpoppins:
I think the title could be worded more clearly. The average user would look at it and think: "Just what exactly does this PR accomplish? Wasn't the purpose of -warp and IDCLEV to allow warping to any available map from day one?". He would be right, too: it was, just with some limitations. And this PR relaxes this limitations, but it's not quite the case of being able to warp to _any_ available map still.

AFAICS, it's MAP00-MAP99 now for D2 (rather than the old MAP01-MAP33) and E1M0-E9M99 for D1 (rather than the old E1M1-E4M9), with zero-numbered maps allowed on all episodes. Is this basically it?

Also, I note that the `(gamemode == commercial && (epsd > 1))` (condensed) check has been removed. I wonder what the implications of this would be.

fabiangreffrath:
> I think the title could be worded more clearly. 

Alright, I guess "extending the range of maps to reach by IDCLEV and -warp" would have been more fitting then. :wink:

> AFAICS, it's MAP00-MAP99 now for D2 (rather than the old MAP01-MAP33) and E1M0-E9M99 for D1 (rather than the old E1M1-E4M9), with zero-numbered maps allowed on all episodes. Is this basically it?

Yep, that's it. Everything beyond that is UMAPINFO territory.

> Also, I note that the `(gamemode == commercial && (epsd > 1))` (condensed) check has been removed. I wonder what the implications of this would be.

The episode for Doom 2 is hard-coded to `1` earlier in that code. This means, unless there is a UMAPINFO entry for the requested map available, you'll have no chance than to warp to an "episode 1" map in Doom 2. And if there is a UMAPINFO entry available, the checks would be skipped anyway. 

Whatever, on second glance, some of the test even make sense, so I'll probably add some back. At least some basic range checks for `epsd` and `map`. As it is now, the code interprets any two subsequent key presses after "IDCLEV" to form a map number.

PR #239, opened 2021-03-13
Mouse stuff

fabiangreffrath:


prboom2/src/SDL/i_video.c
```diff
@@ -1427,6 +1417,24 @@ static void DeactivateMouse(void)
...
+  {
+    int x, y;
+
+    SDL_GetRelativeMouseState(&x, &y);
+
+    if (x != 0 || y != 0)
+    {
+      event_t event;
+      event.type = ev_mouse;
+      event.data1 = 0;
```
kraflab, line 1430: This will turn all the buttons off each frame, right? Or will it? 🤔
I disambiguated also to avoid this problem - but the mouse motion code in g_game.c was already separated from the button checking code in dsda-doom, during the input rewrite, while they are still coupled in pr+.

fabiangreffrath: Yep, something was still odd. I have now separate even types for mouse buttons and motion (taken from dsda, of course) and now things seem to be working alright.

kraflab (approved):

PR #242, opened 2021-03-18
Fix P_SetMobjState stack

kraflab:
If you try to play back Map 31 UV Max from https://dsdarchive.com/wads/mayhem18p you may crash on windows, because we run out of stack space (linux seems to usually have more stack space).

The P_SetMobjState method exhibits recursion and allocates a 4k integer array onto the stack. In the aforementioned demo, the recursion reaches a depth of over 150, which yields 600k integers on the stack from this method alone (it isn't the only method in the chain).

This PR allocates the 4k state tracking table onto the heap instead, which prevents the crash from happening.

### DIFF
--- a/prboom2/src/p_mobj.c
+++ b/prboom2/src/p_mobj.c
@@ -71,10 +71,10 @@ dboolean P_SetMobjState(mobj_t* mobj,statenum_t state)
   static int recursion;                       // detects recursion
   statenum_t i = state;                       // initial state
   dboolean ret = true;                         // return value
-  statenum_t tempstate[NUMSTATES];            // for use with recursion
+  statenum_t* tempstate = NULL;               // for use with recursion
 
   if (recursion++)                            // if recursion detected,
-    memset(seenstate=tempstate,0,sizeof tempstate); // clear state table
+    seenstate = tempstate = calloc(NUMSTATES, sizeof(statenum_t)); // allocate state table
 
   do
     {
@@ -109,6 +109,9 @@ dboolean P_SetMobjState(mobj_t* mobj,statenum_t state)
   if (!--recursion)
     for (;(state=seenstate[i]);i=state-1)
       seenstate[i] = 0;  // killough 4/9/98: erase memory of states
+
+  if (tempstate)
+    free(tempstate);
 
   return ret;
 }


fabiangreffrath:
A stupid question, maybe, but would it have been possible to re-use the array across recursions by declaring it as `static`? Or does each recursion require a clean state? Just asking, didn't investigate the code any further...

alexey-lysiuk:
Did you measure how added allocations affect performance on monster-heavy maps?

From my point of view, an allocation in such function is the last resort, and it must be avoided at all cost.
Certainly, it shouldn't use system's thread safe routines with their noticeable overhead.

I would rather increase the default increase stack size for Windows target.
IIRC, it's 1 MB on Windows while POSIX targets have 8 MB for the main thread.

kraflab:
> A stupid question, maybe, but would it have been possible to re-use the array across recursions by declaring it as `static`? Or does each recursion require a clean state? Just asking, didn't investigate the code any further...

Each recursion uses its own array.

kraflab:
> Did you measure how added allocations affect performance on monster-heavy maps?
> 
> From my point of view, an allocation in such function is the last resort, and it must be avoided at all cost.
> Certainly, it shouldn't use system's thread safe routines with their noticeable overhead.
> 
> I would rather increase the default stack size for Windows target.
> IIRC, it's 1 MB on Windows while POSIX targets have 8 MB for the main thread.

I played back an hour long demo with 1k+ monsters and there was no difference in performance with -nodraw -timedemo.

alexey-lysiuk:
Sorry but "no difference" is quite subjective definition. Performance comparison should not be done like that. Each version needs to be tested several times, and results should be provided. E.g., run old version 3 to 5 times in row, same for new one, and post their output.

kraflab:
> Sorry but "no difference" is quite subjective definition. Performance comparison should not be done like that. Each version needs to be tested several times, and results should be provided. E.g., run old version 3 to 5 times in row, same for new one, and post their output.

I have a background in computational physics and parallel computing and understand how to measure change. There was no statistical difference between the two. Sorry I didn't show my work. 😛

One person's no-change result isn't meaningful, it was just to check if there's a noticeable change, so there's not much point in posting numbers. I'm not about to proactively ask for performance tests from multiple people for every change though - people will complain if there's an issue. If you're worried about it you should try it yourself. Then if you do see an issue on your end we can try a different solution.

Increasing the stack size won't actually fix the problem, just the particular demo, so I don't think that's a good direction to take.

There are plenty of fancy ways to fix this, if there is a problem.

kraflab:
Here's a chart of the crashing demo in question, and the calls to this method. In blue are the number of calls total, and in red are the number of calls that require allocations. Keep in mind the first call in a chain doesn't allocate anything. X axis is leveltime. 597 monsters at start.

chart (2)

The huge spike that caused the crash is due to a quirk with the custom enemies.

Here's the same chart for an eviternity demo (809 monsters at start):

chart (3)

This is to give a sense of how relatively rare recursive calls are. You could probably measure a one tic performance hit on the massive spike in the first demo, but the comparison of a hypothetical performance hit for one frame is against a crash 😛

fabiangreffrath:
Impressive, thank you!

PR #247, opened 2021-03-20
Fix menu intermission desync

kraflab:
If you open the menu during intermission while the numbers are ticking down, you will cause a desync. This was reported some time ago but hadn't been addressed yet. Here is the fix.


fabiangreffrath:
Awesome, thank you!

PR #251, opened 2021-03-22
Fix off-by-one bug (fixes #249)

### LINKED ISSUE #249: map33 still shows map03 as next (bug)
This could be a simple off-by-one oversight.

The recent fix works great for me (thank you) on all >33 mapsbut not for map33 which still shows `map03` as the next level.

Tested it on a 6d06b1bb build.

**Steps**
1. Start DOOM2 with mayhem2020
2. Type `idclev33` then `idclev`
3. Look at the next map

runlow:
32 is 31 in index 0

### DIFF
--- a/prboom2/src/e6y.c
+++ b/prboom2/src/e6y.c
@@ -455,7 +455,7 @@ int G_GotoNextLevel(int *e, int *m)
 		if (gamemode == commercial)
 		{
 			epsd = 1;
-			if (map >= 0 && map <= 32)
+			if (map >= 0 && map <= 31)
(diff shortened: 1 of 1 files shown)


fabiangreffrath:
The MAP33->MAP03 transition is canonical: 
https://doomwiki.org/wiki/MAP33:_Betray_(Xbox_Doom_II)

runlow:
That's a different m5sum though. I don't understand.

fabiangreffrath:
There are more than one official IWADs available that contain MAP33.

PR #253, opened 2021-03-22
Update comp_respawn comments

kraflab:
When this comp option was implemented, there was an oversight in EE that had the meaning of the comp option flipped ("on" should always mean "not fixed"). That was fixed in EE some time in the past though, so comp_respawn has the same meaning now. This PR just updates a couple comments in the code to prevent confusion.

### DIFF
--- a/prboom2/src/doomstat.h
+++ b/prboom2/src/doomstat.h
@@ -109,7 +109,7 @@ enum {
   comp_infcheat,
   comp_zerotags,
   comp_moveblock,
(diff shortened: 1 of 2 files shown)


fabiangreffrath:
Could you name a commit, a version or a date when this fix was fixed in EE? Because now we are inverting the contents of the comments which might lead to additional confusion in the future.

fabiangreffrath:
Whatever, seems like it's at least 15 years ago, i.e. before the current code was checked in into GIT. I was just afraid that inverting the comments commenting the inversion would lead to more confusion. 😵

PR #254, opened 2021-03-23
fix custom episode select for Doom 2

rfomin:
By default, Doom 2 has 4 episodes in the current UMAPINFO implementation. There are few UMAPINFO files out in the wild that don't aware of that.
Before:
doom00
After:
doom00

### DIFF
--- a/prboom2/doc/umapinfo.txt
+++ b/prboom2/doc/umapinfo.txt
@@ -38,7 +38,7 @@ intertextsecret = clear		; Disables default intermission text for the given map'
 interbackdrop = "graphic"	
 							; Backdrop to be used for intertext and intertextsecret. If it does not specify a valid flat, it will be drawn as a patch instead. If not specified the FLOOR4_8 flat will be used, regardless of which map it is used on.
 intermusic = "song"			; Music to be used for intertext and intertextsecret. If not specified D_VICTOR/D_READ_M will be used, depending on the IWAD.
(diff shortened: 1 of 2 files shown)


fabiangreffrath:
Aren't you supposed to explicitly call `episode = clear` in that case?

rfomin:
It's not clear in the specification, also doesn't make sense - Doom 2 doesn't have 4 episodes.

fabiangreffrath:
Makes sense. Could you please clarify the spec and increase the revision log?

fabiangreffrath (approved):
LGTM

fabiangreffrath:
Any objections anyone?

JadingTsunami:
Personally I would instead initialize `EpiDef` correctly based on gamemode/gamemission instead of waiting for a customization to apply the correct values. But this is perhaps just stylistic preference.

Also if we are fixing Doom 2 we should fix Chex Quest also; this would mean checking for `(gamemode == commercial || gamemission == chex)` like is done in `m_menu.c:710`.

fabiangreffrath:
I think @JadingTsunami is right. @rfomin Could you please add chex to the check (and the docs)?

rfomin:
> add chex to the check (and the docs)

Does this mean that every port that implements UMAPINFO must support Chex Quest to some extent?

fabiangreffrath:
> > add chex to the check (and the docs)
> 
> Does this mean that every port that implements UMAPINFO must support Chex Quest to some extent?

I don't think so, but most do anyway. 😉

PR #255, opened 2021-03-23
reformat umapinfo.txt so that it is easier to read in plaintext

runlow:
* hard wrap everything

* comments above the keys

* extra newlines between sections and other minor tweaks


fabiangreffrath (approved):
I like this formatting!

PR #258, opened 2021-03-26
consider translucency and comp_translucency settings on all complevels

### LINKED ISSUE #246: [REQUEST] Allow translucency on all complevels
Shouldn't affect demos, so I'm not sure why this is complevel'd anyway.

Sorry @fabiangreffrath, just saw your request to post this.

### LINKED ISSUE #248: (Feature request) An option to disable translucency in projectiles in software﻿ mode
Title, you can switch to opengl and then switch again to software mode to disable translucency in projectiles somehow tho, but i want this as an option.

fabiangreffrath:
This changes the code to evaluate the user-set
`default_comp[comp_translucency]` value instead of the
`comp[comp_translucency]` value which is computed based on current
complevel.

This allows for translucent sprites on all complevels (fixes #246) and
at the same time allows for disabling predefined translucency on all
complevels (fixes #248). Additionally, the change of the compatibility
option is instantly applied to the current game.

### DIFF
--- a/prboom2/src/d_deh.c
+++ b/prboom2/src/d_deh.c
@@ -1490,7 +1490,7 @@ void deh_changeCompTranslucency(void)
       }
       else
 #endif
-      if (comp[comp_translucency]) 
+      if (default_comp[comp_translucency])
         mobjinfo[predefined_translucency[i]].flags &= ~MF_TRANSLUCENT;
       else 
         mobjinfo[predefined_translucency[i]].flags |= MF_TRANSLUCENT;
--- a/prboom2/src/m_menu.c
+++ b/prboom2/src/m_menu.c
@@ -3686,7 +3686,7 @@ setup_menu_t comp_settings3[] =  // Compatibility Settings screen #3
   {"Max Health in DEH applies only to potions", S_YESNO, m_null, C_X,
    C_Y + compat_maxhealth * COMP_SPC, {"comp_maxhealth"}},
   {"No predefined translucency for some things", S_YESNO, m_null, C_X,
-   C_Y + compat_translucency * COMP_SPC, {"comp_translucency"}},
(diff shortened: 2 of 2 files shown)


fabiangreffrath:
So, this change enables users to bend the speedrunning rules when recording Vanilla complevel demos. It does not mean they will definitely break the rules, though, because they can still disable translucency as a whole. There are other ways to cheat if you are really determined to. It's all a matter of trust anyway, but I understand that "if it's recorded with Pr+ then it's definitely rules compliant" has some value - although I doubt that this is already the case in all depth given how many "visual enhancements" (show alive monsters, health bar above monsters, game speed, etc.) Pr+ does already offer (I understand that these features are still problematic but that the ship has sailed a long time ago).

Anyway, this is about the change at hand. On the hand hand I think that we don't have to adhere to speedrunning rules. Especially, if these rules are more focused on the source port being used rather than their features. For example, Crispy Doom does also allow for translucent objects and colored blood when recording Vanilla complevel demos, but is still considered safe. This is a user-requested feature and I don't see why we should hold the wishes of the speedrunning community higher than those of occasional players or non-speedrunning just-for-fun demo recorders.

On the other hand it would be a pity if this port would lose its rank as the universally recommended go-to port for speedrunning just because it has become possible to configure it in a way that bends the rules. Is it the port's responsibility to make sure its recordings are rules compliant? Isn't this the exact reason why dsda-doom is developed and isn't this about to become the de-facto recommended choice for speedrunning anyway?

I hope you see how torn I am - again. Whatever I decide, one of both communities will be disappointed. :tired_face:

kraflab:
At this point as you've said I think the ship has sailed as far as the recommended speedrunning port: it's basically dsda-doom or crispy, based on a user's preferences for feature set and wads they will run. I think this isn't a bad thing though - it removes the onus of worrying about that kind of thing from mainline pr+ 🤷

Issue #226, opened 2021-02-26
HoM on Plutonia map11 with OpenGL


snoggie1487:
There is a HoM bug at the very start of Plutonia map11. 
![doom02](https://user-images.githubusercontent.com/38010396/109367688-67e02800-785c-11eb-95d5-76e43b6aa6d9.png)
![doom03](https://user-images.githubusercontent.com/38010396/109367689-69115500-785c-11eb-87ee-8f9eb2031164.png)
![doom05](https://user-images.githubusercontent.com/38010396/109367691-69a9eb80-785c-11eb-96cb-5aa54128db15.png)
![doom00](https://user-images.githubusercontent.com/38010396/109367692-6a428200-785c-11eb-9a87-f7867fcc3ded.png)
![doom01](https://user-images.githubusercontent.com/38010396/109367693-6a428200-785c-11eb-994d-5bed32ebfffe.png)

runlow:
Not sure if related and if it isn't the mapmaker's fault - but there's pretty bad HOM in Alien Vendetta map11 when you enter the waterfall.

![2021-02-28-093007_1920x1080_scrot](https://user-images.githubusercontent.com/2176657/109413906-bd751b80-79a7-11eb-9f94-b123a12cb276.png)

![2021-02-28-093000_1920x1080_scrot](https://user-images.githubusercontent.com/2176657/109413901-bb12c180-79a7-11eb-8589-5baf54b53d66.png)

snoggie1487:
This occurs when a sector fully connects with the floor / ceiling (not both) of the surrounding sector, and has no sidedef textures. This was used in Plutonia map11 as an invisible monster-los blocking solid wall. I made a quick map to demonstrate this, try looking at the ground with software rendering and then opengl rendering. 
[opengl_bug.zip](https://github.com/coelckers/prboom-plus/files/6127298/opengl_bug.zip)

fabiangreffrath:
Thank you very much for the test map! I hope this will help somebody nail down this issue down. Unfortunately, this somebody won't be me, I am OpenGL-illiterate. 😞 

However, it raises one question: If the sector's floor is above the view plane (`viewz`), you will be looking beneath the floor, so strictly speaking there can't be a visible floor flat on the ground. Which part of the software renderer decides to continue using the floor flat of the surrounding sector in this case?

JadingTsunami:
I think PrBoom tries to fix this, but it seems like it's broken or maybe just not working in these cases.

[...]

Issue #173, opened 2021-01-18
OpenGL rendering issues


T-Will133:
When using the OpenGL renderer, some textures become misaligned and sectors with F_SKY1 can block sectors behind it if the ceiling is set low enough.

Using E4M6 to demonstrate.

8-bit software
![doom03](https://user-images.githubusercontent.com/59149832/104963992-3a70a680-59a1-11eb-8fbe-b2272ca6899c.png)

OpenGL
![doom02](https://user-images.githubusercontent.com/59149832/104964002-3fcdf100-59a1-11eb-8ead-388c2923d531.png)

Build
64-bit (commit e96bc52)

OS
Windows 10 Home 2004

fabiangreffrath:
This has been reported before:
https://sourceforge.net/p/prboom-plus/bugs/235/

T-Will133:
Since this has been reported before, what is preventing it from being resolved?

JadingTsunami:
It's not a simple fix. I don't think anyone has been interested to try and re-write the parts of the renderer that would need to be re-done to address it.

T-Will133:
Had a feeling it wouldn’t be easy to fix. Still great to see PrBoom being supported on some way.

JadingTsunami:
I thought of a simple way to fix this, but it will need some thought and testing to ensure there are no caveats that cause glitches elsewhere.

The basic idea would be skip generation of blocking sky walls when the front-facing sector is higher than the back-facing sector, and both sectors are sky textured.

I tested and verified this fixes this particular E4M6 issue.

fabiangreffrath:
Reopening as this is currently being worked on.

PR #259, opened 2021-03-28
[Optimized Implementation] Proposal to Fix flood HOM

JadingTsunami:
Gaps in flat rendering cause a HOM void in the GL renderer where the software renderer would bleed through the flat behind the void. This code is an initial attempt to remedy this (see #226). 

There are a few drawbacks:

* It pulls `P_Spec` calls into `R_Bsp` which is not ideal from an encapsulation perspective.
* A known corner case exists if an off-screen control sector is adjacent to the on-screen sector at a lower/higher sector height. This will cause the fake flat to take on the properties of the wrong sector. However, due to the nature of this effect, I have never seen the sector effect used this way and I suspect it may be rare to never.

Comment is invited as I believe there are many ways to fix this, and some may see advantages to one approach over others.

### DIFF
--- a/prboom2/src/gl_missingtexture.c
+++ b/prboom2/src/gl_missingtexture.c
@@ -167,7 +167,11 @@ void gld_PreprocessFakeSectors(void)
   int groupid;
 
   if (gl_use_stencil)
+  {
+    // precalculate NO_TOPTEXTURES and NO_BOTTOMTEXTURES flags
+    gld_PrepareSectorSpecialEffects();
     return;
+  }
 
   // free memory
   if (fakeplanes)
--- a/prboom2/src/gl_missingtexture.c
+++ b/prboom2/src/gl_missingtexture.c
@@ -75,7 +75,7 @@ static void gld_PrepareSectorSpecialEffects(void)
   for (num = 0; num < numsectors; num++)
   {
     // the following is for specialeffects. see r_bsp.c in R_Subsector
-    sectors[num].flags = (NO_TOPTEXTURES | NO_BOTTOMTEXTURES);
+    sectors[num].flags |= (NO_TOPTEXTURES | NO_BOTTOMTEXTURES);
 
     for (i=0; i<sectors[num].linecount; i++)
     {
@@ -95,6 +95,36 @@ static void gld_PrepareSectorSpecialEffects(void)
           sectors[num].flags &= ~NO_TOPTEXTURES;
         if (side1->bottomtexture != NO_TEXTURE)
           sectors[num].flags &= ~NO_BOTTOMTEXTURES;
+
+        /* sides should not have null sectors, but we check anyway */
+        if (side0->sector && side1->sector)
+        {
+            dboolean front_floor_is_sky = (side0->sector->floorpic == skyflatnum);
+            dboolean front_ceil_is_sky = (side0->sector->ceilingpic == skyflatnum);
+
+            dboolean back_floor_is_sky = (side1->sector->floorpic == skyflatnum);
+            dboolean back_ceil_is_sky = (side1->sector->ceilingpic == skyflatnum);
+
+            dboolean needs_back_lower = !(front_floor_is_sky) && (side0->sector->floorheight > side1->sector->floorheight);
+            dboolean needs_front_lower = !(back_floor_is_sky) && (side0->sector->floorheight < side1->sector->floorheight);
+
+            dboolean needs_front_upper = !(back_ceil_is_sky) && (side0->sector->ceilingheight > side1->sector->ceilingheight);
+            dboolean needs_back_upper = !(front_ceil_is_sky) && (side0->sector->ceilingheight < side1->sector->ceilingheight);
+
+            /* now mark the sectors that may require HOM bleed-through */
+            /* TODO: mark "bleed height" here */
+            if (needs_front_upper && side0->toptexture == NO_TEXTURE)
+                side1->sector->flags |= MISSING_TOPTEXTURES;
+            if (needs_back_upper && side1->toptexture == NO_TEXTURE)
+                side0->sector->flags |= MISSING_TOPTEXTURES;
+
+            if (needs_back_lower && side1->bottomtexture == NO_TEXTURE)
+                side0->sector->flags |= MISSING_BOTTOMTEXTURES;
+            if (needs_front_lower && side0->bottomtexture == NO_TEXTURE)
+                side1->sector->flags |= MISSING_BOTTOMTEXTURES;
+
+        }
+
       }
       else
       {
--- a/prboom2/src/gl_missingtexture.c
+++ b/prboom2/src/gl_missingtexture.c
@@ -82,6 +82,14 @@ static void gld_PrepareSectorSpecialEffects(void)
 {
   int i, num;
 
+  /* free memory if allocated by previous maps */
+  if (bleedsectors) {
+      free(bleedsectors);
+      numbleedsectors = 0;
+      bleedsectors = NULL;
+  }
+
+
   for (num = 0; num < numsectors; num++)
   {
     // the following is for specialeffects. see r_bsp.c in R_Subsector
@@ -196,7 +204,9 @@ static void gld_RegisterBleedthroughSector(sector_t* source, sector_t* target, i
      * and register it instead */
 
     if ((bleedsectors[source_idx].target == NULL) ||
-             (bleedsectors[source_idx].target && ((ceiling && bleedsectors[source_idx].target->ceilingheight > target->ceilingheight) || (bleedsectors[source_idx].target->floorheight < target->floorheight))))
+        (bleedsectors[source_idx].target &&
+        ((ceiling && bleedsectors[source_idx].target->ceilingheight > target->ceilingheight) ||
+        (bleedsectors[source_idx].target->floorheight < target->floorheight))))
         bleedsectors[source_idx].target = target;
 }
 
--- a/prboom2/src/gl_missingtexture.c
+++ b/prboom2/src/gl_missingtexture.c
@@ -166,9 +166,6 @@ void gld_PreprocessFakeSectors(void)
   int i, j, k, ceiling;
   int groupid;
 
-  if (gl_use_stencil)
-    return;
-
   // free memory
   if (fakeplanes)
   {
--- a/prboom2/src/gl_missingtexture.c
+++ b/prboom2/src/gl_missingtexture.c
@@ -83,12 +83,12 @@ static void gld_PrepareSectorSpecialEffects(void)
   int i, num;
 
   /* free memory if allocated by previous maps */
-  if (bleedsectors) {
+  if (bleedsectors)
+  {
       free(bleedsectors);
       numbleedsectors = 0;
       bleedsectors = NULL;
   }
-
 
   for (num = 0; num < numsectors; num++)
   {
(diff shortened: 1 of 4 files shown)


fabiangreffrath (commented):
Thank you very much for taking a look at this! I cannot say much about the change itself as I haven't wrapped my head around the OpenGL renderer yet. But the code looks clean and reasonable. 

Remaining questions:
- Does this change also fix the rendering bug in E4M6, i.e. https://github.com/coelckers/prboom-plus/issues/173 ?
- Is it possible to preprocess and cache some of the calls to `P_FindLowestFloorSurrounding()` and `P_FindModelFloorSector()`?
- What exactly does `use_gl_stencil` mean and what happens if this is enabled? Will a different fix be applied? Does the user have any control over this variable?

JadingTsunami:
Thanks for the review! Here are my thoughts...

> * Does this change also fix the rendering bug in E4M6, i.e. #173 ?

Unfortunately no, that one is caused by depth buffering and will require a different fix. Basically, the software renderer has no concept of a depth buffer, but it is an integral part of the OpenGL renderer. To mimic Doom's SW renderer, the GL renderer adds invisible "sky walls" to block the depth buffer (so walls "behind" the sky won't render at all). The problem is that the sky ceiling is *below* the top of the player's view height at that point, and those invisible sky walls block the marble walls from rendering. It's kind of a logical way to render the scene, but the Doom SW renderer does not always follow such logic... :)

> * Is it possible to preprocess and cache some of the calls to `P_FindLowestFloorSurrounding()` and `P_FindModelFloorSector()`?

Yes, in fact this is what the calls to `GetBestFake()` were doing, but I removed them after some consideration. Since it is a rare special effect (maybe 99% of maps do not use it), my reasoning was it's better to take a few extra cycles while rendering
 a rare "special effect" than add preprocessing time to every level.

But, I think perhaps I am thinking about it too deeply. :)

What changed my mind was, I would need to re-write parts of `GetBestFake` and add some new data structures. And adding more data structures, tracking and other stuff for this effect did not seem worth the tradeoff overall to me. But, if that is the preferred approach, I don't mind that way either.

> * What exactly does `use_gl_stencil` mean and what happens if this is enabled? Will a different fix be applied? Does the user have any control over this variable?

Yes, this can be configured in the main config file. It's whether or not the user's hardware supports stenciling. I think virtually all hardware supports it at this point. The stencil if check is actually legacy code that I left untouched.

In the SW renderer, if a sector has no bottom or top textures and is a raised floor or lowered ceiling, PrBoom will attempt to "bleed" that sector over to create the same effect as the SW renderer in such a case. There are two ways to do this: one that uses openGL stencils and one that does not. The code there is choosing the approach based on that support. The user could disable this in the config file, but I don't see why they would ever do this as the stencil approach should suffice.

kraflab:
I'd vote to keep GetBestFake and make the changes there - at least, to see how they look.

JadingTsunami:
> I'd vote to keep GetBestFake and make the changes there - at least, to see how they look.

Sure. I will mark this as WIP and give some time to optimize it.

JadingTsunami:
Here's an updated example of what I'd do for preprocessing vs. calling p_spec in the renderer, with the new structure I envisioned. This also has a slight advantage that if the map author uses a control sector and correctly adds upper/lower textures, it should be handled correctly.

prboom2/src/r_bsp.c
```diff
@@ -40,6 +40,7 @@
 #include "r_bsp.h" // cph - sanity checking
 #include "v_video.h"
 #include "lprintf.h"
+#include "p_spec.h"
```
fabiangreffrath, line 43: Do we still need this here?

prboom2/src/gl_missingtexture.c
```diff
@@ -111,6 +159,56 @@ static void gld_PrepareSectorSpecialEffects(void)
...
+
+    bleedsectors[source_idx].source = source;
+    bleedsectors[source_idx].ceiling = ceiling;
+
+    /* either register the proposed target since it is first,
+     * or check if the new proposed target is a better option
+     * and register it instead */
+
+    if ((bleedsectors[source_idx].target == NULL) ||
+             (bleedsectors[source_idx].target && ((ceiling && bleedsectors[source_idx].target->ceilingheight > target->ceilingheight) || (bleedsectors[source_idx].target->floorheight < target->floorheight))))
```
fabiangreffrath, line 199: Please break this line.

prboom2/src/gl_missingtexture.c
```diff
@@ -111,6 +159,56 @@ static void gld_PrepareSectorSpecialEffects(void)
...
+        bleedsectors[source_idx].target = target;
+}
+
+sector_t* GetBestBleedSector(sector_t* source, int ceiling)
+{
+    int i;
+    for (i = 0; i < numbleedsectors; i++)
+        if (bleedsectors[i].source == source && bleedsectors[i].ceiling == ceiling)
```
fabiangreffrath, line 207: Why only ceiling?

prboom2/src/r_bsp.c
```diff
@@ -664,39 +665,90 @@ static void R_Subsector(int num)
...
+                    {
+                        dummyceilingplane.height = tmpsec->ceilingheight;
+                        dummyceilingplane.lightlevel = tmpsec->lightlevel;
+                        dummyceilingplane.picnum = tmpsec->ceilingpic;
+                        ceilingplane = &dummyceilingplane;
+                    }
+                }
+            }
+
+            sector_t* tmpsec = NULL;
```
fabiangreffrath, line 705: I think you could get away with only one tmpsec declaration at this point.

prboom2/src/gl_missingtexture.c
```diff
@@ -111,6 +159,56 @@ static void gld_PrepareSectorSpecialEffects(void)
...
+    assert(target);
+
+    /* check whether the sector is processed already */
+    for (i = 0; i < numbleedsectors && source_idx == -1; i++)
+        if (bleedsectors[i].source == source && bleedsectors[i].ceiling == ceiling)
+            source_idx = i;
+    
+    if (source_idx == -1) {
+        /* allocate memory for new sector */
+        if (!bleedsectors) {
```
fabiangreffrath, line 176: There's an easier more generic solution that considers the fact that `realloc(NULL, ...)` is equal to `malloc(...)`. But that's mostly a matter of style.

prboom2/src/gl_missingtexture.c
```diff
@@ -61,12 +61,22 @@ typedef struct
...
+    sector_t *source; // The sector to receive a fake bleed-through flat
+    sector_t *target; // The floor sector whose properties should be copied
+    int ceiling;      // Ceiling or floor
+} bleedthrough_t;
+
 static int numfakeplanes = 0;
 static fakegroup_t *fakeplanes = NULL;
 static sector_t **sectors2 = NULL;
+static bleedthrough_t *bleedsectors = NULL;
+static int numbleedsectors = 0;
```
fabiangreffrath, line 75: Where is this ever reset? What happens when entering the next map?

prboom2/src/r_bsp.c
JadingTsunami: No, and I thought of the same later on too. :)

prboom2/src/gl_missingtexture.c
JadingTsunami: Ceiling is a flag with encoding 0=floor, 1=ceiling; this is not my convention as I followed the existing `GetBestFake` implementation.

We could modify both to a better or more explicit indication, or I could use a different convention if it is preferred.

prboom2/src/r_bsp.c
JadingTsunami: Yes, it's nicer this way and does not create strange scoping. Fixed.

prboom2/src/gl_missingtexture.c
JadingTsunami: Good catch, I needed the reset at the top of that loop. Fixed.

prboom2/src/gl_missingtexture.c
JadingTsunami: Ah, yes. I thought this might be nicer for readability but I can do it that way too.

fabiangreffrath: I see it the other way round, I find it easier to read if this just uses plain `realloc()`.

prboom2/src/gl_missingtexture.c
```diff
@@ -61,21 +61,39 @@ typedef struct
...
   int i, num;
 
+  /* free memory if allocated by previous maps */
+  if (bleedsectors) {
+      free(bleedsectors);
+      numbleedsectors = 0;
+      bleedsectors = NULL;
+  }
+
+
```
fabiangreffrath, line 92: Excess blank line.

prboom2/src/gl_missingtexture.c
```diff
@@ -95,6 +113,44 @@ static void gld_PrepareSectorSpecialEffects(void)
...
+                side0->sector->flags |= MISSING_BOTTOMTEXTURES;
+                gld_RegisterBleedthroughSector(side0->sector,side1->sector,0);
+            }
+
+            if (needs_front_lower && side0->bottomtexture == NO_TEXTURE) {
+                side1->sector->flags |= MISSING_BOTTOMTEXTURES;
+                gld_RegisterBleedthroughSector(side1->sector,side0->sector,0);
+            }
+        }
+
```
fabiangreffrath, line 153: Excess blank line.

prboom2/src/gl_missingtexture.c
```diff
@@ -111,6 +167,58 @@ static void gld_PrepareSectorSpecialEffects(void)
...
+    /* either register the proposed target since it is first,
+     * or check if the new proposed target is a better option
+     * and register it instead */
+
+    if ((bleedsectors[source_idx].target == NULL) ||
+        (bleedsectors[source_idx].target &&
+        ((ceiling && bleedsectors[source_idx].target->ceilingheight > target->ceilingheight) ||
+        (bleedsectors[source_idx].target->floorheight < target->floorheight))))
+        bleedsectors[source_idx].target = target;
```
fabiangreffrath, line 210: This is still hard to read, could you please add braces?

fabiangreffrath (changes requested):
Alright, this is mostly about style consistency now, but we are near the finishing line.

prboom2/src/gl_missingtexture.c
JadingTsunami: Sure, addressed this in the latest push.

prboom2/src/gl_missingtexture.c
JadingTsunami: Removed excess blank line.

prboom2/src/gl_missingtexture.c
JadingTsunami: Removed excess blank line.

prboom2/src/gl_missingtexture.c
JadingTsunami: I added braces and expanded the logic in keeping with other files (gl_main.c, am_map.c). See if this is new formatting is preferable.

fabiangreffrath (approved):

fabiangreffrath:
Is this still WIP as the title suggests?

JadingTsunami:
> Is this still WIP as the title suggests?

I don't have any more work planned on the implementation, but if anyone has feedback I'm happy to incorporate it. I left it as "WIP" in case there was still more to discuss now that everything is updated.

But, I've updated the title to show I am done with the implementation side.

PR #260, opened 2021-03-30
Fix shifting automap markers

### LINKED ISSUE #90: Automap markers move incorrectly in 16:9 ratio

TwelveEyes:
The drawn patch for the automap marker number always assumed `patch_stretch_full` as the stretch setting, so on `patch_stretch_16x10` or `patch_stretch_4x3` the marker position was not being set correctly. This should correct the behavior for opengl & software, all resolutions, aspect ratios, and stretch settings. Unfortunately I was unable to test every possibility, so there may still be unintended behavior.

Fixes #90

### DIFF
--- a/prboom2/src/am_map.c
+++ b/prboom2/src/am_map.c
@@ -2142,22 +2142,22 @@ static void AM_drawMarks(void)
               if (am_frame.precise)
               {
                 V_DrawNamePatchPrecise(
-                  (float)p.fx * 320.0f / SCREENWIDTH, (float)p.fy * 200.0f / SCREENHEIGHT,
-                  FB, namebuf, CR_DEFAULT, VPT_STRETCH);
+                  (float)p.fx / patches_scalex, (float)p.fy * 200.0f / SCREENHEIGHT,
+                  FB, namebuf, CR_DEFAULT, VPT_ALIGN_LEFT | VPT_STRETCH);
               }
               else
               {
                 V_DrawNamePatch(
-                  p.x * 320 / SCREENWIDTH, p.y * 200 / SCREENHEIGHT,
-                  FB, namebuf, CR_DEFAULT, VPT_STRETCH);
+                  p.x / patches_scalex, p.y * 200 / SCREENHEIGHT,
+                  FB, namebuf, CR_DEFAULT, VPT_ALIGN_LEFT | VPT_STRETCH);
               }
             }
 
             w += markpoints[i].widths[k] + 1;
-            p.x += w * SCREENWIDTH / 320;
+            p.x += w * patches_scalex;
             if (am_frame.precise)
             {
-              p.fx += (float)w * SCREENWIDTH / 320.0f;
+              p.fx += (float)w * patches_scalex;
             }
           }
         }
--- a/prboom2/src/am_map.c
+++ b/prboom2/src/am_map.c
@@ -2185,30 +2185,10 @@ static void AM_drawMarks(void)
           }
 
           w += markpoints[i].widths[k] + 1;
-
-          switch (render_stretch_hud)
+          p.x += w * SCREENWIDTH / 320;
+          if (am_frame.precise)
           {
-            case patch_stretch_16x10:
-              p.x += w * patches_scalex;
-              if (am_frame.precise)
-              {
-                p.fx += (float)w * patches_scalex;
-              }
-              break;
-            case patch_stretch_4x3:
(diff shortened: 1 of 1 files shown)


fabiangreffrath (changes requested):
Cool, thanks! It's just a matter of style, but since only the x and y coordinates and the flags change between the three possible `render_stretch_hud` values, you could only adjust these in the `switch` block and then call the patch drawing function based `am_frame.precise` - just as it was before. This will make the code a bit easier to read and highlight the actual differences in the arguments.

TwelveEyes:
Did you mean like this?

```diff
--- a/prboom2/src/am_map.c
+++ b/prboom2/src/am_map.c
@@ -2137,17 +2137,47 @@ static void AM_drawMarks(void)
           if (p.x < f_x + f_w &&
               p.x + markpoints[i].widths[k] * SCREENWIDTH / 320 >= f_x)
           {
+            float fx, fy;
+            int x, y, flags;
+
+            switch (render_stretch_hud)
+            {
+              case patch_stretch_16x10:
+                fx = (float)p.fx / patches_scalex;
+                fy = (float)p.fy * 200.0f / SCREENHEIGHT;
+
+                x = p.x / patches_scalex;
+                y = p.y * 200 / SCREENHEIGHT;
+
+                flags = VPT_ALIGN_LEFT | VPT_STRETCH;
+                break;
+              case patch_stretch_4x3:
+                fx = (float)p.fx * 320.0f / WIDE_SCREENWIDTH;
+                fy = (float)p.fy * 200.0f / WIDE_SCREENHEIGHT;
+
+                x = p.x * 320 / WIDE_SCREENWIDTH;
+                y = p.y * 200 / WIDE_SCREENHEIGHT;
+                
+                flags = VPT_ALIGN_LEFT | VPT_STRETCH;
+                break;
+              case patch_stretch_full:
+                fx = (float)p.fx * 320.0f / SCREENWIDTH;
+                fy = (float)p.fy * 200.0f / SCREENHEIGHT;
+
+                x = p.x * 320 / SCREENWIDTH;
+                y = p.y * 200 / SCREENHEIGHT;
+                
+                flags = VPT_ALIGN_WIDE | VPT_STRETCH;
+                break;
+            }
+
             if (am_frame.precise)
             {
-              V_DrawNamePatchPrecise(
-                (float)p.fx * 320.0f / SCREENWIDTH, (float)p.fy * 200.0f / SCREENHEIGHT,
-                FB, namebuf, CR_DEFAULT, VPT_ALIGN_WIDE | VPT_STRETCH);
+              V_DrawNamePatchPrecise(fx, fy, FB, namebuf, CR_DEFAULT, flags);
             else
             {
-              V_DrawNamePatch(
-                p.x * 320 / SCREENWIDTH, p.y * 200 / SCREENHEIGHT,
-                FB, namebuf, CR_DEFAULT, VPT_ALIGN_WIDE | VPT_STRETCH);
+              V_DrawNamePatch(x, y, FB, namebuf, CR_DEFAULT, flags);
             }
           }
```

fabiangreffrath:
Beautiful.

TwelveEyes:
Alright, requested changes have been made.

fabiangreffrath (approved):
I had no chance to check if this fixes the issue for all possible values of `render_stretch_hud`, but code wise it looks fine.

fabiangreffrath:
All seems fine, thanks!

PR #261, opened 2021-04-01
Fix sky walls for cases with sky vs. sky toptextures like E4M6.

JadingTsunami:
Proposed fix for #173.

I had some time to look into this bug, and it's simpler than I anticipated. The code I had planned to add is already there. The issue is that the code that adds depth buffer blocking sky walls is checking for height differences with upper textures but doesn't verify the upper texture would be applied to a sky sector. The Doom software renderer does not render upper textures against the sky. This PR checks that top textures applied to a sky sector do not affect the decision to generate blocking sky walls.

### DIFF
--- a/prboom2/src/gl_main.c
+++ b/prboom2/src/gl_main.c
@@ -1747,7 +1747,7 @@ void gld_AddWall(seg_t *seg)
         }
         else
         {
-          if ((toptexture != NO_TEXTURE && midtexture == NO_TEXTURE) ||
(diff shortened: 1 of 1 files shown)


fabiangreffrath:
Is there some pendant code for bottom textures that needs fixing as well?

JadingTsunami:
> Is there some pendant code for bottom textures that needs fixing as well?

Strangely enough, the Doom SW renderer (and in turn  the OpenGL renderer) respects bottomtextures even on sky flats.

fabiangreffrath (approved):

Issue #112, opened 2020-07-26
GLB+ cannot sort sprites and walls simultaneously - causes issue w/boom translucent things


physixcat:
I have noticed that using the OpenGL renderer results in Boom's translucent things to render as solid. I investigated the reason for this and found an old thread from 2012 that @fabiangreffrath made: https://sourceforge.net/p/prboom-plus/bugs/218/

According to Blondie: 

> _"Boom's translucent things (imp fireballs, plasma, teleporters, etc.) are **permanently disabled** at all complevels in OpenGL mode because **GLBoom cannot sort sprites and walls at the same time**."_

 Fortunately, Blondie provides a DEH patch that can force translucency to be enabled/disabled, but I suspect this is just supposed to be a bandaid solution.

Upon being asked about how to solve the issue, Blondie replied 

> _"You'll have to ask Andrey for the specifics. I reported the "smooth sprite edges" bug years ago to confirm if it was another manifestation of the sorting issue, and indeed it was. I also requested a config variable years ago to **toggle GL sprite translucency**, but it was never implemented. Doesn't matter, though, because those dehacked patches work just as well."_ 

I thought I'd bring the topic up again, in hopes that someone has an idea of how to solve the root cause

JadingTsunami:
Does anyone have a picture of the effect (expected vs. actual) here? I am finding it hard to visualize what this means.

I don't think there should be any issue rendering projectiles as translucent, but I haven't looked into this issue yet.

PR #262, opened 2021-04-01
Fix transparent sprites in OpenGL mode

JadingTsunami:
Proposed fix for #112.

I would like to open the discussion on resolving transparent sprites and walls.

The old PrBoom renderer placed all draw items into a single array and sorted them by distance. This method had many drawbacks. PrBoom+ splits the render items up according to type, which is much cleaner, but introduces a problem for transparent items in differing categories.

Sorting the two simultaneously is a fairly big effort, I think. There is a much simpler approach of determining which element is farther and rendering that one first. This is almost an "in-place sort" by popping items off the tail of the array and comparing them.

There is a bit of awkward code needed to unpack the x/y coordinates of the GLSprite's original Thing. The coordinates are packed into an integer, presumably to simplify sorting, and must be unpacked to check which seg side the original Thing is on when comparing a transparent wall (seg) vs. a transparent Thing.

I think the speedup from merging and sorting the transparent items in the same array would not be so great as it only saves the unpack and comparison operations during the render stages.

### DIFF
--- a/prboom2/src/gl_main.c
+++ b/prboom2/src/gl_main.c
@@ -2624,6 +2624,8 @@ void gld_ProjectSprite(mobj_t* thing, int lightlevel)
 
   sprite.index = gl_spriteindex++;
   sprite.xy = thing->x + (thing->y >> 16); 
+  sprite.fx = thing->x;
+  sprite.fy = thing->y;
 
   sprite.vt = 0.0f;
   sprite.vb = sprite.gltexture->scaleyfac;
@@ -3259,8 +3261,8 @@ void gld_DrawScene(player_t *player)
               int ti;
               for (ti = tsprite_idx; ti >= 0; ti--) {
                   /* reconstruct the sprite xy */
-                  fixed_t tsx = gld_drawinfo.items[GLDIT_TSPRITE][ti].item.sprite->xy & ~(0x0000FFFF);
-                  fixed_t tsy = (gld_drawinfo.items[GLDIT_TSPRITE][ti].item.sprite->xy & ~(0xFFFF0000)) << FRACBITS;
+                  fixed_t tsx = gld_drawinfo.items[GLDIT_TSPRITE][ti].item.sprite->fx;
+                  fixed_t tsy = gld_drawinfo.items[GLDIT_TSPRITE][ti].item.sprite->fy;
 
                   if (R_PointOnSegSide(tsx, tsy, twseg))
                   {
--- a/prboom2/src/gl_main.c
+++ b/prboom2/src/gl_main.c
@@ -3240,11 +3240,31 @@ void gld_DrawScene(player_t *player)
 
       while (twall_idx >= 0 || tsprite_idx >= 0 )
       {
-          /* find out which is next to draw */
-          GLTexture *tw = (twall_idx >= 0) ? gld_drawinfo.items[GLDIT_TWALL][twall_idx].item.wall->gltexture : NULL;
-          GLTexture *ts = (tsprite_idx >= 0) ? gld_drawinfo.items[GLDIT_TSPRITE][tsprite_idx].item.sprite->gltexture : NULL;
+          dboolean draw_tsprite = true;
 
-          if (((tw && ts) && (ts >= tw)) || (ts && !tw))
+          /* find out which is next to draw */
+          if (twall_idx >= 0 && tsprite_idx >= 0)
+          {
+              /* both are left to draw, determine
+               * which is farther */
+              seg_t *twseg = gld_drawinfo.items[GLDIT_TWALL][twall_idx].item.wall->seg;
+              /* reconstruct the sprite xy */
+              fixed_t tsx = gld_drawinfo.items[GLDIT_TSPRITE][tsprite_idx].item.sprite->xy & 0xFFFF0000;
+              fixed_t tsy = (gld_drawinfo.items[GLDIT_TSPRITE][tsprite_idx].item.sprite->xy & 0x0000FFFF) << 16;
+
+              if (R_PointOnSegSide(tsx, tsy, twseg))
+              {
+                  draw_tsprite = true;
+              }
+          }
+          else if (twall_idx >= 0)
+          {
+              /* no transparent sprites left, draw a wall */
+              draw_tsprite = false;
+          }
+          /* fall-through case is draw sprite */
+
+          if (draw_tsprite)
           {
               /* transparent sprite is farther, draw it */
               glAlphaFunc(GL_GEQUAL, gl_mask_sprite_threshold_f);
@@ -3255,7 +3275,7 @@ void gld_DrawScene(player_t *player)
               glDepthMask(GL_TRUE);
               glAlphaFunc(GL_GEQUAL, 0.5f);
           }
-          else if (((tw && ts) && (tw >= ts)) || (tw && !ts))
+          else
           {
               /* transparent wall is farther, draw it */
               glAlphaFunc(GL_GREATER, 0.0f);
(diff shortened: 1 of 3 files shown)


fabiangreffrath:
Here is a little test map:

[test2.zip](https://github.com/coelckers/prboom-plus/files/6246307/test2.zip)

JadingTsunami:
> Here is a little test map:
> [test3.zip](https://github.com/coelckers/prboom-plus/files/6246401/test3.zip)

Thanks! Looks to be working now to me.

fabiangreffrath:
Yes, it seems to be working, awesome! However, it might be due to my eyes, though, I find it hard to tell which sphere is in front of the wall and which is behind. To me it looks like they are all drawn as if they were behind the wall.

prboom2/src/gl_main.c
```diff
@@ -3232,41 +3232,59 @@ void gld_DrawScene(player_t *player)
...
+          {
+              /* both are left to draw, determine
+               * which is farther */
+              seg_t *twseg = gld_drawinfo.items[GLDIT_TWALL][twall_idx].item.wall->seg;
+              /* reconstruct the sprite xy */
+              fixed_t tsx = gld_drawinfo.items[GLDIT_TSPRITE][tsprite_idx].item.sprite->xy & 0xFFFF0000;
+              fixed_t tsy = (gld_drawinfo.items[GLDIT_TSPRITE][tsprite_idx].item.sprite->xy & 0x0000FFFF) << 16;
+
+              if (R_PointOnSegSide(tsx, tsy, twseg))
+              {
```
fabiangreffrath, line 3256: Is it true that this rather expensive calculation is only ever executed if there are both translucent walls and sprites in the rendered scene? And, now comes the obvious question, is it somehow possible to cache this information from an earlier stage of the rendering process? I mean, the software renderer also knows if the sprite is in front of the wall or behind it, but this is probably unrelated.

JadingTsunami: Correct - this code is only executed if both translucent walls and sprites (things) are present in the scene at the same time.

To cache the relative distances would require depth sorting of both object types based on a common reference point. I think it would require essentially the same calculations. I am not sure it would save too much in the end.

JadingTsunami: > Yes, it seems to be working, awesome! However, it might be due to my eyes, though, I find it hard to tell which sphere is in front of the wall and which is behind. To me it looks like they are all drawn as if they were behind the wall.

Ah, I think you are right - the logic is reversed in that case.

Oh there is one limitation I should mention -- the overlap is determined based on the center of the thing and not the true sprite edges. So when a rendered sprite is partially inside a translucent wall, it may not render as partially in and out of the wall at the same time and there is a "slicing" effect. This is perhaps something I can put as a future fix option.

JadingTsunami:
Move to WIP to give me some time to optimize the sorting outside the "critical region" of the renderer.

JadingTsunami:
Some discussion will be needed here, as there are tradeoffs.

The simplest solution I see for pre-caching the render order would be to iterate through each transparent thing and trace a line back to the player, and check if any transparent lines (segs) are crossed along the way. Based on that, assign transparent sprites and flats to groups that are rendered together in order (farthest to nearest).

The limitation is this only considers the center of the sprite and not the rendered boundaries, so in cases of partial overlap, rendering may be incorrect.

Otherwise I think it will be much more computation intensive and would have to happen in the renderer.

But I am interested to hear what others think too.

fabiangreffrath:
Your proposed solution for pre-caching sounds a lot more complicated than the current approach. It also sounds more computationally expensive, as each sprite needs to be checked against each seg, which - at least as I understand it - isn't strictly necessary currently. 

The software renderer sorts sprites and segs by their scale factor, but that's probably unrelated here. 

Regarding the overlap of sprite edges and segs, I'd consider this a luxury problem for now, given that it is currently impossible to have them rendered in the same scene at all.

JadingTsunami:
> Your proposed solution for pre-caching sounds a lot more complicated than the current approach. It also sounds more computationally expensive, as each sprite needs to be checked against each seg, which - at least as I understand it - isn't strictly necessary currently.

Yes, this is correct. As I see it there are two reasonable algorithms we can choose from, or three if we are willing to trade speed for accuracy. And I will set aside the problem of partial overlap since, as you pointed out, that's a bit of a luxury fix at the moment. In other words, sprites are considered only from their center point and not from the edges.

Option 1:

* Assign each transparent seg an order grouping value, from farthest to nearest. Each transparent sprite now can also have an order grouping relative to a seg.
* For each transparent sprite (farthest to nearest), draw a line from the player to the sprite.
* For each transparent seg (farthest to nearest), check if the line drawn in step 1 intersects the seg.
  * If yes for any seg, order the sprite in the nearest group "farther" than the seg and stop.
  * If no for all segs, the sprite will render in normal sprite order.

Now, there are corner cases that Option 1 can't solve without checking every transparent seg. But I think for most cases this would not happen. For a solution to all cases, each transparent seg has to be checked and the distance from the seg intercept to the player has to be calculated and then compared, and the longest distance used to determine render order.

I think this option is robust but maybe not worth all the compute power it would consume. I wouldn't want transparent lines and transparent things in the same scene to render significantly differently than an ordinary scene, and this computation is all compute-bound and not graphics-bound (unfortunately).

Option 2:

Do what's done in the current patch, possibly with one exception:

There is a "corner case" the current patch doesn't handle. The function `PointOnSegSide` only checks on which side of the seg a point lies and not whether the seg and point actually overlap. Thus the following case fails:

```
        1*
    -----T
2*
         <-P-
```

`1*` and `2*` are transparent things, the `-T` line is a transparent line, and the P arrow is the player looking left. The player should see transparent thing `1*` rendered behind the transparent line. But because `2*` is farther and doesn't block the seg, the seg is rendered before `1*` causing it to render on the wrong side. To address this we would need to cycle through all remaining transparent things and see if the next seg in the render order is blocked by any of them.

In such a case we have solved almost all the cases, I think. With multiple overlapping transparent segs there can be odd cases, but they should be very rare I think.

Whichever is preferable I think can be implemented. I have the 2nd option working already, so if that option is acceptable I can upload it.

fabiangreffrath:
I am all for option 2, let's see what you got.

JadingTsunami:
Sounds good, I have updated the branch.

Please note it is possible to break this simplified algorithm as I mentioned, and with many transparent walls and transparent sprites on screen at once, I think it is possible that some will render in front or behind others incorrectly at certain angles. If the sprites are projectiles, this will be momentary in all likelihood.

If there are only sprites on screen, or only transparent walls, there should be no issues. Small or transient visual glitches are possible with both overlapping, due to the nature of the simplification here.

But as you said, right now it is broken, so I suppose it is an enhancement to have them working right now.

fabiangreffrath:
Sorry, but this is looking broken now. If you load my test map, step back until you reach the wall and then turn and strafe a bit, the spheres will begin to flicker between being drawn in front of and behind the wall. Also, all spheres will flicker whenever the Imp throws a fireball.

JadingTsunami:
> Sorry, but this is looking broken now. If you load my test map, step back until you reach the wall and then turn and strafe a bit, the spheres will begin to flicker between being drawn in front of and behind the wall. Also, all spheres will flicker whenever the Imp throws a fireball.

Yes, these are the inaccuracies with the approach. Your test map has transparent things overlapping (same x,y coordinate) with a transparent seg which this won't handle very well. Also I don't believe `R_PointOnSegSide` is accurate at short distances. I believe we'd need to move away from using `R_PointOnSegSide` and toward the approach I laid out in option 1 to avoid situations like this.

But in case others have some ideas: The basic problem is, given a set of sorted sprites and segs, how do you choose them in exact far-to-near order? The approach here is to try to take the farthest in each set and compare them.

fabiangreffrath:
Wow, what an improvement this detail made!

JadingTsunami:
> Wow, what an improvement this detail made!

Yes, I think this is a much better compromise. You may notice the "slicing" effect I mentioned for the transparent sprites that are embedded in the transparent seg (overlapping x/y coordinate) in your test map. But I think perhaps this is not so critical to fix with urgency.

fabiangreffrath:
Yes, that's a minor detail that might need some fix in the future but shouldn't hold this implementation back. Let *the perfect* not become the enemy of *the good*!

fabiangreffrath (approved):

PR #269, opened 2021-04-06
generate a default save slot name if the user saves to an empty slot

fabiangreffrath:
or if the save slot name already starts with a map identifier


JadingTsunami:
This is a nice feature! Looks good to me.

Issue #206, opened 2021-02-16
Fix sky in Eviternity.wad (opengl render)


JuanchoES:
The sky looks wrong, but looks correct in sofware mode

fabiangreffrath:
Which map do you mean? IIRC Eviternity has a custom sky for about every map. What do you mean with "wrong" and "correct", please elaborate. How about screenshots?

PhoxFyre:
Also which version are you using? As you have "(Glboom+)" in your issue title, I think you may be using 2.5.1.4 or 2.5.1.5.test which have the glboom-plus.exe included, whereas the latest here (2.6um) only has prboom-plus.exe included and internal setting to change to OpenGL. This may be a tangential aside, but please make sure you are using the current most up-to-date version as 2.5.1.4 and 2.5.1.5.test are not within the realms of the development of this repo.

JuanchoES:
im using this version https://github.com/coelckers/prboom-plus/releases/tag/v2.6um sorry for the video quality 


https://user-images.githubusercontent.com/70743149/108202442-678bb280-7121-11eb-9428-c41def18dfcb.mp4

jackrjli:
In your config file, what is the value of gl_skymode? Try setting that to 3 if it isn't.

JuanchoES:
> In your config file, what is the value of gl_skymode? Try setting that to 3 if it isn't.

It works! Should i keep this setting forever? Whats the differences between the other settings of "gl_skymode"? Thx
![image](https://user-images.githubusercontent.com/70743149/108262581-65e3de00-7165-11eb-8ce6-69a2729c1ede.png)

jackrjli:
I don't know what exactly gl_skymode does, but I haven't run into any problems keeping it on 3. That said, it doesn't solve _all_ problems the GL renderer has with skies (for example, the skies in Ancient Aliens are misaligned no matter what skymode you use, most noticeably in the last few maps).

fabiangreffrath:
`gl_skymode` being set to `0`, which is the default, means `skytype_auto` [1], which in turn means `skytype_skydome` if mouselook is enabled or `skytype_standard` if it isn't [2]. Setting `gl_skymode` to `3` permanently sets it to `skytype_skydome`. Don't ask me what the difference is, though...


[1] https://github.com/coelckers/prboom-plus/blob/995b06347f6db3745fe877f5d826571435f03db1/prboom2/src/gl_struct.h#L52-L60
[...]

PR #273, opened 2021-04-09
Fix sky scaling for non-standard sky sizes.

JadingTsunami:
Proposed fix for #206.

When rendering a the sky as a dome in OpenGL rendering mode, the y-scaling is calculated based on a fixed factor of 200/160. This change would take the sky texture's vertical size into account.

From my local testing it resolves the issues with non-standard sky sizes such as Eviternity and Ancient Aliens.


fabiangreffrath (approved):

fabiangreffrath:
Thank you!

PR #274, opened 2021-04-10
Health bar updates

jengelh:


prboom2/src/gl_main.c
```diff
@@ -2376,31 +2381,17 @@ static void gld_DrawHealthBars(void)
     for (i = count - 1; i >= 0; i--)
     {
       GLHealthBar *hbar = gld_drawinfo.items[GLDIT_HBAR][i].item.hbar;
-      if (hbar->cm != cm)
-      {
-        cm = hbar->cm;
-        glColor4f(cm2RGB[cm][0], cm2RGB[cm][1], cm2RGB[cm][2], 1.0f);
-      }
+      glColor4f(hbar->r, hbar->g, 0, 1);
```
JadingTsunami, line 2384: Floats should be passed here, not integers, even for fixed values.

prboom2/src/gl_main.c
```diff
@@ -2328,20 +2328,26 @@ static void gld_DrawSprite(GLSprite *sprite)
...
+    } else if (health_percent <= health_bar_yellow) {
+      hbar.r = 1;
+      hbar.g = (double)(health_percent - health_bar_red) /
+               (health_bar_yellow - health_bar_red);
+    } else if (health_percent <= health_bar_green) {
+      /* FullSimplify[1-((percent-yellow)/(green-yellow))] */
+      hbar.r = (double)(health_bar_green - health_percent) /
+               (health_bar_green - health_bar_yellow);
+      hbar.g = 1;
+    }
```
JadingTsunami, line 2349: Passes uninitialized values when the mobj health is 100%.

Following the logic here, there should be an else case with r=0;g=1.

Ex:

```c
 else {
      hbar.r = 0;
      hbar.g = 1;
    }
```

prboom2/src/gl_main.c
```diff
@@ -2328,20 +2328,26 @@ static void gld_DrawSprite(GLSprite *sprite)
...
 
-    if (hbar.cm >= 0)
     {
```
JadingTsunami, line 2351: Is this in a code block on purpose?

JadingTsunami:
Left some review comments. I also noticed this change causes health bars to always appear, rather than only show up when the monster or thing is damaged. I am not sure if this was intentional or not.

I will leave the question of whether this change is desired or not to others.

prboom2/src/gl_main.c
jengelh: The compiler would implicitly convert them, especially so for constant expressions. Are you getting a compiler warning or something that would support passing the exact type from the get-go?

prboom2/src/gl_main.c
jengelh: I did this intentionally, yes. By showing the health bar at all times, one becomes aware that something is shootable in the first place, like customized decoration. Hence the (hbar.cm >= 0) condition was removed. As new variables are declared and C89 is picky about mixing declarations and code, I had just left the block there to get a minimal patch. I can certainly add a patch to remedy that.

Having bars everywhere is not so much a detriment; in case of e.g. nuts.wad, most monsters will get damaged anyway so that even the original health bar's implementation would readily show across the rendered scene.

fabiangreffrath (changes requested):
Sorry, no. In its current form this patch is much too invasive for the little change that you want to achieve. No coding style changes or "I would have done it that way" kind of changes,  please. Also, please don't change numerical values that people got used to over the past two decades.

fabiangreffrath:
Additionally, I don't think that every barrel should have a health bar - always visible - or every shootable fake window in Hacx or whatever. This changes a bit too much without good reason. 

I think the health bars were introduced for speed runners to give them a hint how far they are from 100% kills on the map.

jengelh:
> I think the health bars were introduced for speed runners to give them a hint how far they are from 100% kills on the map.

That makes no sense to me. To see if you have 100% kills on the map, there's the STS part of the HUD.

jengelh:
The current 6566b04 proposal should have it much closer to the original now. With a blank prboom-plus.cfg, bars only get shown for objects that have taken damage.

prboom2/src/gl_main.c
```diff
@@ -2328,20 +2328,29 @@ static void gld_DrawSprite(GLSprite *sprite)
 
 static void gld_AddHealthBar(mobj_t* thing, GLSprite *sprite)
 {
-  if (((thing->flags & (MF_COUNTKILL | MF_CORPSE)) == MF_COUNTKILL) && (thing->health > 0))
+  if ((thing->flags & MF_SHOOTABLE) && thing->info->spawnhealth > 0 &&
```
fabiangreffrath, line 2331: Still not convinced of this change, though. I wouldn't want health bars for barrels.

fabiangreffrath, line 2331: Still not convinced of this change, though. I wouldn't want health bars for barrels.

prboom2/src/gl_main.c
```diff
@@ -2328,20 +2328,29 @@ static void gld_DrawSprite(GLSprite *sprite)
...
-      hbar.cm = CR_YELLOW;
-    else if (health_percent <= health_bar_green)
-      hbar.cm = CR_GREEN;
+    hbar.r = hbar.g = -1.0f;
+    if (health_percent <= 0) {
+      hbar.r = 1.0f;
+      hbar.g = 0.0f;
+    } else if (health_percent <= health_bar_red) {
+      hbar.r = 1.0f;
+      hbar.g = (double)health_percent / (health_bar_yellow - health_bar_red);
```
fabiangreffrath, line 2343: Why not cast to `float` here?

prboom2/src/gl_main.c
```diff
@@ -2328,20 +2328,29 @@ static void gld_DrawSprite(GLSprite *sprite)
...
+    } else if (health_percent <= health_bar_yellow) {
+      hbar.r = (double)(health_bar_yellow - health_percent) /
+               (health_bar_yellow - health_bar_red);
+      hbar.g = 1.0f;
+    } else if (health_percent <= health_bar_green) {
```
fabiangreffrath, line 2348: If this applies to all shootable objects, it applies to other players as well, who might have more health than `spawnhealth`.

fabiangreffrath (changes requested):
Much better now, thanks, though there are still some details left to discuss.

prboom2/src/gl_main.c
jengelh: Knowing the barrel's state is just as essential as knowing a monster's health state; a barrel can usually take a lowpower hit, sometimes two. This influences whether it's "reasonably safe" to speedrun past it. Or, strategy: tickling the barrel raises the chance it will definitely go off on the next hit, and if that final blow is dealt by a monster, that's free infighting for the speedrunner.

kraflab: I wouldn't expect to see health bars on barrels either.

prboom2/src/gl_main.c
```diff
@@ -2328,20 +2328,29 @@ static void gld_DrawSprite(GLSprite *sprite)
...
     int health_percent = thing->health * 100 / thing->info->spawnhealth;
 
-    hbar.cm = -1;
-    if (health_percent <= health_bar_red)
-      hbar.cm = CR_RED;
-    else if (health_percent <= health_bar_yellow)
-      hbar.cm = CR_YELLOW;
-    else if (health_percent <= health_bar_green)
-      hbar.cm = CR_GREEN;
+    hbar.r = hbar.g = -1.0f;
```
JadingTsunami, line 2337: Hmm, I don't think initializing an illegal value is appropriate here, especially when you only check for one illegal value later. While it works today, it could easily break in the future.

I still recommend having a catch-all else case that will initialize the values to something legal but innocuous, and `r=0.0f;g=1.0f;` seems to fit to me.

Others may have differing opinions.

kraflab:
Is there some clamouring for this change? It doesn't appear to fix anything, and I personally don't like the visual changes either 🤷

fabiangreffrath:
I could live with the color gradient but the "all shootable objects" change is a blocker.

prboom2/src/gl_main.c
JadingTsunami: Oh, need to guard against divide-by-zero here as well. The cutoff values are user-input and are not guarded or bounds-checked. Certain settings would crash the game every time a health bar appeared.

jengelh:
Added a config flag for whether it's shown for MF_SHOOTABLEs or the traditional MF_COUNTKILL.

prboom2/src/gl_main.c
jengelh: added a config option for this, `health_bar_shootables`.

prboom2/src/gl_main.c
jengelh: done

prboom2/src/gl_main.c
jengelh: I can't see anything negative stemming from a mobj having health_percent>100.

prboom2/src/gl_main.c
jengelh: Fine, I'll test for both.

prboom2/src/gl_main.c
jengelh: If you managed to set yellow<red in the .cfg file, the block will not be executed because it is always testing for <=red first.

prboom2/src/gl_main.c
```diff
@@ -2326,22 +2326,38 @@ static void gld_DrawSprite(GLSprite *sprite)
...
+    if (health_percent <= 0) {
+      hbar.r = 1.0f;
+      hbar.g = 0.0f;
+    } else if (health_percent <= health_bar_red) {
+      hbar.r = 1.0f;
+      hbar.g = yr == 0 ? 0 : (float)health_percent / yr;
```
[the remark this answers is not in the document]
jengelh: It does not.

prboom2/src/gl_main.c
```diff
@@ -2326,22 +2326,38 @@ static void gld_DrawSprite(GLSprite *sprite)
   }
 }
 
+static int gld_EvaluateShowBar(mobj_t* thing)
```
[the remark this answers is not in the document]
jengelh: Hence it behind behind a config optionn.

PR #276, opened 2021-04-10
Update .gitignore

jengelh:
The build leaves a number of new files lying around.
prboom.wad no longer exists, prboom-plus.wad has moved, etc.


alexey-lysiuk (commented):
Do an out-of-tree build, and you won't have any of those files.

fabiangreffrath:
Right, but if some of the files in `.gitignore` are plain moot, it won't hurt to update that file.

fabiangreffrath (approved):

fabiangreffrath:
Thank you!

PR #277, opened 2021-04-10
New advanced HUD layout

Fireb3rt:
doom00


fabiangreffrath (approved):

fabiangreffrath:
I see you added this as the 13th HUD - which is good. Or, is this rather supposed to replace an existing HUD?

I mean, we cannot collect good HUDs ad ultimo. On the other hand, we shouldn't probably repalce existing HUDs that some users got used to...?

Fireb3rt:
This is supposed to be a new HUD.

fabiangreffrath:
Alright, thank you! One last thing: There is the newline byte missing at the end of the file. This isn't fatal, but ugly. Most probably you edited the file with Windows' own `notepad.exe`, right? Could you probably open it up again with a *real* editor, e.g. notepad++, hit Enter and then Backspace at the end of the file and save it again?

Fireb3rt:
Sorry, no idea what I'm doing. Hopefully it's good now.

fabiangreffrath:
Sorry I confused you. Yes, it's perfect now.

PR #278, opened 2021-04-14
Support widescreen assets

### LINKED ISSUE #142: [Suggestion] Support for widescreen graphics
Since the Unity port, Crispy Doom and GZDoom have Widescreen graphics support I would like to see Prboom+ get the same treatment, maybe something could be done with [Widepix](https://forum.zdoom.org/viewtopic.php?f=46&t=67843) as well but repackaged into .wad formats.

fabiangreffrath:
Fixes #142

### DIFF
--- a/prboom2/src/st_stuff.c
+++ b/prboom2/src/st_stuff.c
@@ -56,6 +56,7 @@
 int ST_SCALED_HEIGHT;
 int ST_SCALED_WIDTH;
 int ST_SCALED_Y;
+int ST_SCALED_OFFSETX;
 
 // Palette indices.
 // For damage/bonus red-/gold-shifts
@@ -311,7 +312,7 @@ static patchnum_t faces[ST_NUMFACES];
 static patchnum_t faceback; // CPhipps - single background, translated for different players
 
 //e6y: status bar background
-static patchnum_t stbarbg;
+static patchnum_t stbarbg = {0};
 patchnum_t grnrock;
 patchnum_t brdr_t, brdr_b, brdr_l, brdr_r;
 patchnum_t brdr_tl, brdr_tr, brdr_bl, brdr_br;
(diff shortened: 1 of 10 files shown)


fabiangreffrath:
Ready for review!

UnluckySpade7:
Hi. I've been doing some testing with the unity 1.3 iwad and found a few things that might need fixing such as:

The status bar not rendering properly (mostly on the right side) and leavinng a black background.
Unitybar


Unity port comparison:
BarComparison

And the bunny scroller cutting off near the end:
Bunnyscroll

fabiangreffrath:
@UnluckySpade7 Thank you for checking out this branch! It seems that both issues are caused by improper calculation of the patch boundaries outside the [0..320] interval here: 
https://github.com/coelckers/prboom-plus/blob/3419bfbc6411c5c02116b521450033e9e263f442/prboom2/src/v_video.c#L558

fabiangreffrath:
Could you please check if your issues are resolved now?

UnluckySpade7:
yep, everything works fine, thanks.

kraflab:
Seems like e3 intermission is busted (this is WidePix, 1280x720).
doom16

fabiangreffrath:
Does this probably need an offset?
https://github.com/coelckers/prboom-plus/blob/adbf50015b737b9e5f7e5ebf5473477b935b7743/prboom2/src/wi_stuff.c#L758

prboom2/src/st_stuff.c
```diff
@@ -421,6 +421,8 @@ void ST_SetScaledWidth(void)
       break;
   }
 
+  ST_SCALED_WIDTH = (ST_SCALED_WIDTH + 3) & (int)~3;
```
kraflab, line 424: If the bar isn't a multiple of 4 in width, then this rounds up. If it rounds up to the screenwidth, then the offset will be 0. Then the side-filling code thinks "there's nothing to fill" and does a no-op. But actually there are <4 pixels empty. Then if you're in opengl it applies palette alpha into the empty region, but doesn't refresh the area when drawing the bar because it "isn't there", so the color gets stuck until another palette effect.

Is it an engine limitation that it has to be 4?

fabiangreffrath: I guess, technically it needs to be an even number, because it is later subtracted from another even number (i. e. `SCREENWIDTH`) and the difference divided by two to achieve `ST_SCALED_OFFSETX`:
```
ST_SCALED_OFFSETX = (SCREENWIDTH - ST_SCALED_WIDTH) / 2;
```
Obviously, you'll want `(SCREENWIDTH - ST_SCALED_WIDTH)` to be even, but for some reason I wanted `ST_SCALED_OFFSETX` to be an even number as well. I cannot remember the exact reason, though.

PR #280, opened 2021-04-16
Add DEHEXTRA/Crispy Doom actor names to ActorNames[], for BossAction

Shadow-Hog:
Exactly what it says on the tin. I'd like to use these new actor definitions as makeshift TIDs in my project, so I'm adding in the missing names to this array (sourced from ZDoom's names for DECORATE/ZScript objects, for consistency).

[Attached a sample WAD](https://github.com/coelckers/prboom-plus/files/6323020/TestingExtraThings2.zip) consisting of a BFG, ample plasma ammo, and way too many Former Humans, who are actually custom monsters in the final DEHEXTRA definition slot. To open the door forward, you must kill all the DEHEXTRA monsters. A vanilla Former Human is also in the map behind the door that opens for comparison (which you also must kill to open the second and  final door to the exit). This should prove that the additions to this list work.

One note - my comments note that the Crispy _Doom_ custom blood colors, rocket trails and head-pop gib objects are unused, but I'm not certain that's the case (though I believe they aren't). If I'm wrong, I can adjust the comments to _not_ make that claim.

(Deleted and re-forked my repo, so hopefully this won't be as much of a headache as the DEHEXTRA sounds were.)


fabiangreffrath:
Could you please document this in the spec and also add a changelog entry?

Shadow-Hog:
See no reason why not, I'll look into doing so after work.

prboom2/src/umapinfo.cpp
```diff
@@ -191,6 +191,114 @@ static const char * const ActorNames[] =
 	"PlasmaBall2",
 	"EvilSceptre",
 	"UnholyBible",
+	// Crispy/Retro
+	"MusicChanger",
```
kraflab, line 195: There shouldn't be anything port-specific in this file I think. Incidentally this entity is in pr+ too, so I don't think the crispy / retro comment makes sense.

MT_GIBDTH, which is 145, is used in pr+ as well.

The other comments about blood and rocket trail I would remove as well...these have entries in pr+'s info.h related to stealth creatures for instance, they have no defined cross-port meaning, or meaning in the standard then as far as I understand it.

Shadow-Hog: Comments have been removed, save the one noting where the 100 DEHEXTRA slots begin.

prboom2/doc/umapinfo.txt
```diff
@@ -307,10 +307,119 @@ Thingtypes:
     PlasmaBall2
     EvilSceptre
     UnholyBible
+    MusicChanger
+    Deh_Actor_145
```
fabiangreffrath, line 311: Sorry for nagging you again, but I guess for documentation purposes it will be sufficient to abbreviate these with 
```
Deh_Actor_145
...
Deh_Actor_249
```

Shadow-Hog: Done!

fabiangreffrath (approved):

Issue #279, opened 2021-04-15
Music tracks assigned with UMAPINFO reset when a map resets.


TrilbyTrillion:
Music tracks assigned with UMAPINFO reset when a map resets, such as with a death. 

map E4M1
{
   music = "D_E4M1"
}

This only happens with tracks reassigned with this method and not with the usual method of replacing existing tracks.  A minor glitch that doesn't break anything, but felt it was worth bringing attention to.

fabiangreffrath:
Isn't this a duplicate of https://github.com/coelckers/prboom-plus/issues/102?

TrilbyTrillion:
Oh, Sorry, I hadn't seen that. I had reported it over on the Doomworld forum and was told to submit an issue here.

JadingTsunami:
I wasn't aware of the duplicate issue and mentioned that @TrilbyTrillion should file an issue here. Apologies for encouraging the dupe. It does look like the same issue and the fix from Crispy Doom seems appropriate to me.

fabiangreffrath:
Could anyone provide a PR for this please?

JadingTsunami:
Sure, as requested: #281

Issue #102, opened 2020-07-02
Music restarting when loading a save or restarting in a map with UMAPINFO-designated music


Alper002:
Method of reproducing this issue:
Step 1:Get a wad file with a UMAPINFO lump, I.E. [DBP25](https://www.doomworld.com/forum/topic/115453-dbp25-dead-but-dreaming/).
Step 2:Enter a map that was defined via UMAPINFO(in DBP25's case, maps 1-11 and map 33)
Step 3:Save
Step 4:Load save

Alternate step 3/4:warp to the map using the idclev cheat

It can get pretty annoying hearing the start of a music track over and over when you're playing a map, rather than hearing the whole of it as you retry and/or load saves.

zstephens:
The same thing happens with musinfo music (a longstanding bug that I'd been meaning to open an issue here about), and it was fixed in crispy doom. perhaps porting that over here would solve this as well?

Here's the crispy commit: https://github.com/fabiangreffrath/crispy-doom/commit/b1825efe185c761f6bf82c2c88a64c50dcec8ebf

Here's some old discussion from when I had attempted to fix it myself: https://github.com/zstephens/prb-plus2515-test/commit/8eb14370bb624a1789d27ee8efecb05b3af1149e

PR #281, opened 2021-04-17
Fix for UMAPINFO music restarting issue.

JadingTsunami:
Proposed fix for #279, #102.

Background information on why this issue happens:

1. `S_ParseMusInfo` wipes out the `musinfo` structure, presuming it will not be used if a `MUSINFO` lump is not present. But since the `UMAPINFO` code re-uses the `musinfo` structure, it should not be wiped out.
2. `S_ChangeMusic` doesn't reset the entries in `S_music` that might be in use by `MUSINFO`. Presumably the parsing took care of this previously. I believe this to be an oversight, but others may be able to say more.

This PR does two things:

1. In `S_ParseMusInfo`, only wipe out the `musinfo` structure if a `MUSINFO` lump is present.
2. `S_ChangeMusic` cleans up `S_music`.

Feedback as always is welcome.

### DIFF
--- a/prboom2/src/s_advsound.c
+++ b/prboom2/src/s_advsound.c
@@ -64,11 +64,22 @@ void S_ParseMusInfo(const char *mapid)
   {
     int num, lumpnum;
     int inMap = false;
+    int load_muslump = -1;
+    /* musinfo item zero is initialized
+     * before we reach the parser and it must be
+     * saved and restored */
+    int itemzero = musinfo.items[0];
+
+    /* don't restart music that is already playing */
+    if (mus_playing &&
+        mus_playing->lumpnum == S_music[NUMMUSIC].lumpnum) {
+        load_muslump = S_music[NUMMUSIC].lumpnum;
+    }
 
     memset(&musinfo, 0, sizeof(musinfo));
-    musinfo.current_item = -1;
-
-    S_music[NUMMUSIC].lumpnum = -1;
+    musinfo.items[0] = itemzero;
+    musinfo.current_item = load_muslump;
+    S_music[NUMMUSIC].lumpnum = load_muslump;
 
     SC_OpenLump("MUSINFO");
 
--- a/prboom2/src/s_advsound.c
+++ b/prboom2/src/s_advsound.c
@@ -59,15 +59,16 @@ musinfo_t musinfo;
 //
 void S_ParseMusInfo(const char *mapid)
 {
-  memset(&musinfo, 0, sizeof(musinfo));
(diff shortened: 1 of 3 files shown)


fabiangreffrath:
Thank you! Does this also fix the music restarting during a save/load cycle?

JadingTsunami:
> Thank you! Does this also fix the music restarting during a save/load cycle?

Yes, in my testing it resolves that issue as well.

prboom2/src/s_advsound.c
```diff
@@ -59,16 +59,17 @@ musinfo_t musinfo;
...
-  musinfo.current_item = -1;
-
-  S_music[NUMMUSIC].lumpnum = -1;
 
   if (W_CheckNumForName("MUSINFO") != -1)
   {
     int num, lumpnum;
     int inMap = false;
 
+    memset(&musinfo, 0, sizeof(musinfo));
```
fabiangreffrath, line 68: Hm, so this only resets the musinfo state if a MUSINFO lump is available. I am afraid that this breaks if you have a map with both UMAPINFO defined music and MUSINFO defined music changers on the map. But it's just a feeling...

JadingTsunami: Hmm, this is a fair point, because UMAPINFO re-uses the `MUSINFO` lump functions to change music, if they are both present, the "music restarting" UMAPINFO bug will reappear. But I am not sure if we can properly fix this without disassociating the two.

Let me take another look and see if perhaps we can keep both...

JadingTsunami:
Here is an updated version. It turns out MUSINFO had some bugs to fix anyway.

A summary of how the execution path works today:

1. `S_Start()` is called and `musinfo` is cleared and preloaded with the map's default music.
2. `S_ParseMusInfo` is called and `musinfo` is cleared again which drops the saved default music from `S_Start`.

This means that the map's default music couldn't be reloaded using MapThing 14100 in any case.

To fix all these issues:

* When calling `S_ChangeMusInfoMusic`, the `UMAPINFO` handler now saves the music lump to `musinfo.items[0]`.
* When `S_ParseMusInfo` clears the `musinfo` structure it will save and restore item 0 (which in either case has been preloaded with the map's default music).
* `S_ParseMusInfo` will also save and restore the playing music, if it was started by a call to `S_ChangeMusInfoMusic`.
    * Note: this required making the `mus_playing` structure visible to other units.

fabiangreffrath:
LGTM, thanks!

fabiangreffrath (approved):

PR #287, opened 2021-04-22
fix endbunny and endgame UMAPINFO fields

### LINKED ISSUE #286: 'endbunny' and 'endgame' fields don't work properly (Doom 1, UMAPINFO)
Test WAD: [endbunnyD1.zip](https://github.com/coelckers/prboom-plus/files/6353146/endbunnyD1.zip)
Exit on E1M2 should end the game and show the bunny scroller. Exit on E1M3 should end the episode.

rfomin:
Fixes #286

### DIFF
--- a/prboom2/src/g_game.c
+++ b/prboom2/src/g_game.c
@@ -1630,7 +1630,7 @@ void G_DoCompleted (void)
   if (gamemapinfo)
   {
 	  const char *next = "";
-	  if (gamemapinfo->endpic[0] && (strcmp(gamemapinfo->endpic, "-") != 0) && gamemapinfo->nointermission)
+	  if (gamemapinfo->endpic[0] && (strcmp(gamemapinfo->endpic, "-") != 0) && !gamemapinfo->nointermission)
 	  {
 		  gameaction = ga_victory;
 		  return;
--- a/prboom2/src/g_game.c
+++ b/prboom2/src/g_game.c
@@ -1630,7 +1630,7 @@ void G_DoCompleted (void)
   if (gamemapinfo)
   {
 	  const char *next = "";
-	  if (gamemapinfo->endpic[0] && (strcmp(gamemapinfo->endpic, "-") != 0) && !gamemapinfo->nointermission)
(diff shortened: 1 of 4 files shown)


prboom2/src/umapinfo.cpp
```diff
@@ -630,7 +630,6 @@ int ParseUMapInfo(const unsigned char *buffer, size_t length, umapinfo_errorfunc
 		if (parsed.endpic[0] && (strcmp(parsed.endpic, "-") != 0))
 		{
 			parsed.nextmap[0] = parsed.nextsecret[0] = 0;
-			if (parsed.endpic[0] == '!') parsed.endpic[0] = 0;
```
JadingTsunami, line 633: Isn't the purpose of this line to handle `endgame = true`? If it is removed, how does that option work?

JadingTsunami:
Looks good to me.

prboom2/src/umapinfo.cpp
fabiangreffrath, line 633: Why was this sucker line there in the first place?

rfomin: Eh, I don't know, just a bug.

prboom2/src/g_game.c
```diff
@@ -1630,7 +1630,7 @@ void G_DoCompleted (void)
   if (gamemapinfo)
   {
 	  const char *next = "";
-	  if (gamemapinfo->endpic[0] && (strcmp(gamemapinfo->endpic, "-") != 0) && gamemapinfo->nointermission)
+	  if (gamemapinfo->endpic[0] && (strcmp(gamemapinfo->endpic, "-") != 0) && !gamemapinfo->nointermission)
```
fabiangreffrath, line 1633: Wait a minute, I just skimmed through this code yesterday. Why didn't I catch this?

This isn't directly related to the original problem in #286, right?

To me this reads like "if we have set an endpic and nointermission is not set, we should skip the intermission screen" which just doesn't sound right.

rfomin: Yes, you are right, I missed one important part of the Woof UMPAINFO implementation and got confused.

fabiangreffrath:
Thank you!

PR #289, opened 2021-04-22
Reset mlook on recording

kraflab:
In pr+, mouselook information is tracked over the course of a demo recording and then appended to the bottom of a demo if it is nonzero for any tick. That tracking doesn't know about demo restarts though, so the buffer fills indefinitely in long recording sessions - and the whole buffer is put into the demo footer when the player exits. So, a demo will contain bad mouselook info when using the restart feature, if mouselook is nonzero for any tick of any demo attempt.

This PR resets the mouselook tick to 0 when beginning a recording, so that it stays consistent with the demo state.

Note that the mouselook data is tracked even if mouselook is on, since it can be turned on or off in the middle of a demo, so the mouselook buffer does affect all users, regardless of whether or not mouselook is turned on.


fabiangreffrath:
Thank you!

Issue #288, opened 2021-04-22
Vanilla fuzz effect in OpenGL renderer


skyrain128:
Could you make an option to use vanilla fuzz effect in OpenGL renderer?

JadingTsunami:
The exact vanilla fuzz effect can't realistically be replicated in OpenGL.

But a "shadow" style function is certainly doable. See below for a simple example of what that would look like, which I put together in the OpenGL renderer just now as an example.

![image](https://user-images.githubusercontent.com/7446082/115795660-2c0d9e80-a385-11eb-889a-f3bd60fbe2ee.png)

fabiangreffrath:
Erm, wow!

fabiangreffrath:
What exactly is the difference to the current implementation?

JadingTsunami:
> What exactly is the difference to the current implementation?

Today, the GL renderer draws a darkened, transparent version of the original sprite when rendering a partially invisible sprite.

The implementation above discards the original sprite color data and instead draws a darkened "shadow" where the sprite would be.

This looks close to the vanilla renderer without the "fuzzing" effect. Recreating the vanilla fuzzing effect would require a new fragment shader and would affect performance. In fact I would recommend against it as I don't see an efficient way to recreate the effect as it is in vanilla without serious performance implications, especially with more than a few partially invisible sprites on screen at once.

Another option could be to require a custom fuzz texture and do an additional rendering pass using the stencil buffer. But this too would hit performance as we'd need to clear the stencil buffer each time, for each transparent sprite. Ideally we'd do one stencil pass but this is not possible if we have multiple sprites on-screen.

I think the above is a fair compromise for those who want a more vanilla look. But I welcome suggestions also.

kraflab:
Definitely digging the shadow effect as a good middleground 👍

fabiangreffrath:
Could you please post a PR for this?

JadingTsunami:
Sure, I will spend some time on this soon.

fabiangreffrath:
Although we still don't have the original Vanilla Doom fuzz effect on the OpenGL renderer, I think this has been sufficiently improved by https://github.com/coelckers/prboom-plus/commit/15431b3a9aac0800c72276aa8dabb13a8f12632e.

PR #292, opened 2021-04-25
Add OpenGL sprite fuzz options.

JadingTsunami:
PR as requested for #288.

Adds three sprite fuzzing (partial invisibility) options, split for player's weapons and things (monsters, etc.):

1. Darken (default): This is the default PrBoom effect which darkens the source sprite and makes it transparent.
2. Shadow: Draw a shadow where the sprite would be, similar to the vanilla fuzz effect without the noise patterns.
3. Transparent: Draw a transparent version of the sprite.

### DIFF
--- a/prboom2/src/gl_main.c
+++ b/prboom2/src/gl_main.c
@@ -156,6 +156,23 @@ GLfloat cm2RGB[CR_LIMIT + 1][4] =
   {1.00f ,1.00f, 1.00f, 1.00f}, //CR_LIMIT
 };
 
+static void gld_SetSpriteFuzzBlendFunc(spritefuzzmode_t fuzzmode)
+{
+    switch (fuzzmode)
+    {
+        default:
+        case fuzz_darken:
+            glBlendFunc(GL_DST_COLOR, GL_ONE_MINUS_SRC_ALPHA);
+            break;
+        case fuzz_shadow:
+            glBlendFunc(GL_ZERO, GL_ONE_MINUS_SRC_ALPHA);
+            break;
+        case fuzz_transparent:
+            glBlendFunc(GL_ONE, GL_ONE_MINUS_SRC_ALPHA);
+            break;
+    }
+}
+
 void SetFrameTextureMode(void)
 {
 #ifdef USE_FBO_TECHNIQUE
@@ -892,19 +909,7 @@ void gld_DrawWeapon(int weaponlump, vissprite_t *vis, int lightlevel)
   // when invisibility is about to go
   if (/*(viewplayer->mo->flags & MF_SHADOW) && */!vis->colormap)
   {
-    switch (gl_weaponspritefuzzmode)
-    {
-        default:
-        case fuzz_darken:
-            glBlendFunc(GL_DST_COLOR, GL_ONE_MINUS_SRC_ALPHA);
-            break;
-        case fuzz_shadow:
-            glBlendFunc(GL_ZERO, GL_ONE_MINUS_SRC_ALPHA);
-            break;
-        case fuzz_transparent:
-            glBlendFunc(GL_ONE, GL_ONE_MINUS_SRC_ALPHA);
-            break;
-    }
+    gld_SetSpriteFuzzBlendFunc(gl_weaponspritefuzzmode);
     glAlphaFunc(GL_GEQUAL,0.1f);
     //glColor4f(0.2f,0.2f,0.2f,(float)tran_filter_pct/100.0f);
     glColor4f(0.2f,0.2f,0.2f,0.33f);
@@ -2268,19 +2273,7 @@ static void gld_DrawSprite(GLSprite *sprite)
     {
       glGetIntegerv(GL_BLEND_SRC, &blend_src);
       glGetIntegerv(GL_BLEND_DST, &blend_dst);
-      switch (gl_thingspritefuzzmode)
-      {
-          default:
-          case fuzz_darken:
-              glBlendFunc(GL_DST_COLOR, GL_ONE_MINUS_SRC_ALPHA);
-              break;
-          case fuzz_shadow:
-              glBlendFunc(GL_ZERO, GL_ONE_MINUS_SRC_ALPHA);
-              break;
-          case fuzz_transparent:
-              glBlendFunc(GL_ONE, GL_ONE_MINUS_SRC_ALPHA);
-              break;
-      }
+      gld_SetSpriteFuzzBlendFunc(gl_thingspritefuzzmode);
       //glColor4f(0.2f,0.2f,0.2f,(float)tran_filter_pct/100.0f);
       glAlphaFunc(GL_GEQUAL,0.1f);
       glColor4f(0.2f,0.2f,0.2f,0.33f);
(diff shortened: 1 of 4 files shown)


prboom2/src/gl_main.c
```diff
@@ -127,6 +127,9 @@ float gl_sprite_offset;       // precalcilated float value for gl_sprite_offset_
 int gl_sprite_blend;  // e6y: smooth sprite edges
 int gl_mask_sprite_threshold;
 float gl_mask_sprite_threshold_f;
+spritefuzzmode_t gl_thingspritefuzzmode;
+spritefuzzmode_t gl_weaponspritefuzzmode;
+const char *gl_spritefuzzmodes[] = {"darken", "shadow", "transparent"};
```
fabiangreffrath, line 132: I think we really only need the array of setting names in m_menu.c?

prboom2/src/gl_main.c
```diff
@@ -2253,7 +2272,23 @@ static void gld_DrawSprite(GLSprite *sprite)
     {
       glGetIntegerv(GL_BLEND_SRC, &blend_src);
       glGetIntegerv(GL_BLEND_DST, &blend_dst);
-      glBlendFunc(GL_DST_COLOR, GL_ONE_MINUS_SRC_ALPHA);
+      switch (gl_thingspritefuzzmode)
+      {
```
fabiangreffrath, line 2276: Do you think we could achieve the same with function pointers instead of checking the same condition for each sprite during each frame?

prboom2/src/gl_main.c
```diff
@@ -2253,7 +2272,23 @@ static void gld_DrawSprite(GLSprite *sprite)
...
+          case fuzz_darken:
+              glBlendFunc(GL_DST_COLOR, GL_ONE_MINUS_SRC_ALPHA);
+              break;
+          case fuzz_shadow:
+              glBlendFunc(GL_ZERO, GL_ONE_MINUS_SRC_ALPHA);
+              break;
+          case fuzz_transparent:
+              glBlendFunc(GL_ONE, GL_ONE_MINUS_SRC_ALPHA);
+              break;
+          default:
```
fabiangreffrath, line 2286: I don't think this is possible to happen. Config values should get normalized upon loading the config file and it's impossible to set them to invalid values in-game. To silence the compiler I would just merge thd default rule and the fuzz_darken case.

fabiangreffrath (changes requested):

prboom2/src/gl_main.c
JadingTsunami: I considered it, but did not do it for the reason that, really the entire thing should be "functionized" at that point:

```c
gl_SaveBlendSettings();
gl_SetSpriteFuzzMode(gl_{whatever}spritefuzzmode);
/* ... */
gl_RestoreBlendSettings();
```

And rather than do such surgery I went with the duplication.

If it is preferred another way, I am of course open to it. I can simply move the blend function calls to a separate function instead. But I think at some point it should get cleaned up, because there are other transparency modes that are not in use that will require more save/restore settings, and if those get used later on, it will require untangling what is done here.

prboom2/src/gl_main.c
JadingTsunami: Fair enough, I was being overprotective in case of future bugs. But merging is a reasonable option too so I will implement that.

prboom2/src/gl_main.c
JadingTsunami: True, I just copied what was done above, trying to follow the convention...if there is such a thing. :)

Line 121 above has the names there. Is it preferable to move these new settings directly into `m_menu.c`, even though it will separate them from the others?

prboom2/src/gl_main.c
JadingTsunami: FYI: The latest branch has the blend functions in a call. I think for now this should suffice, though note that in the future if we add more blending options (say with new shaders or such), this would need to be revisited.

prboom2/src/gl_main.c
```diff
@@ -153,6 +156,23 @@ GLfloat cm2RGB[CR_LIMIT + 1][4] =
   {1.00f ,1.00f, 1.00f, 1.00f}, //CR_LIMIT
 };
 
+static void gld_SetSpriteFuzzBlendFunc(spritefuzzmode_t fuzzmode)
+{
+    switch (fuzzmode)
```
kraflab, line 161: What about `glBlendFunc(x, GL_ONE_MINUS_SRC_ALPHA);` to replace the body of this method? The value `x` either being set when the fuzz mode changes or an array that takes the fuzz mode as its index, depending on what is more convenient?

kraflab: At that point this method probably doesn't need to exist then.

JadingTsunami: That's a reasonable implementation option too.

The argument to keep it in a function would be that, if we decide to expand the fuzzing effects in the future, a function is almost certainly going to be required to do that (more than just setting the blend function). In that case, we have less work to do to expand the function to configure other settings, beyond just the blending function.

For example, there's an option to add a blending equation too that make other invisibility effects possible, such as an "inverted ghost" look. That could can be added to the function but can't be added to `glBlendFunc` directly. Of course this is just an example.

From my side I don't have any concern about either approach; whichever is preferred is fine with me.

prboom2/src/gl_main.c
fabiangreffrath: I guess it's okay like this if it has been done before. I was just wondering why an array that is only used in the menu is defined elsewhere and then made available to the menu only through a header.

prboom2/src/gl_main.c
fabiangreffrath: I can't say what is "preferred" - I can just leave suggestions. 😉 However, given that more implementations for the fuzzing emulation in GL mode are rather unlikely to be added, I agree with @kraflab that it would be neat to have the three possible values for the first argument to `glBlendFunc()` arranged in an array and use `fuzzmode` as an index. This would leave us without any additional branching in this implementation.

prboom2/src/gl_main.c
JadingTsunami: Yes, I agree it probably should be refactored. Maybe a new PR to clean this up at some point, if it becomes necessary.

prboom2/src/gl_main.c
JadingTsunami: Also fine with me. It doesn't really make sense to split up the arguments to `glBlendFunc` so we will need two indexed arrays. I pushed a version that looks like this, and also shows an example of a new option that slightly "brightens" the sprites during blending to show how both `glBlendFunc` factors can be used.

There are many different effects possible, so at some point we may need to bring the function back, but for now I think this is fine also.

fabiangreffrath (approved):
I like this!

fabiangreffrath:
So, is this ready for merging?

JadingTsunami:
> So, is this ready for merging?

Assuming there are no other feedbacks for me to address, it is ready in my view.

kraflab (approved):
👍 thanks for all your work on these topics!

fabiangreffrath:
Thank you!

PR #293, opened 2021-04-26
allow colored blood to get set/overridden by DEHACKED

fabiangreffrath:
This introduces the .blood field in mobjinfo[] which points to a
(virtual) blood mapthing type. By default, point this to MT_BLUEBLOOD
for Cacodemons and MT_GREENBLOOD for Hell nobles and MT_BLOOD for
anything else.

While at it, also introduce the MT_FUZZYBLOOD type for spectre blood,
but do not set it for spectres by default (yet?).

Two questions left:
- Should we initialize to MT_BLOOD for all other monsters or rather
  MT_NULL? This would leave the possibility to apply a special third
  colorization with a crafted RED color translation table.
- Should we unapply the blood assignment if we detect that a monster
  has been dehacked?


kraflab:
We should try to be as consistent as possible across ports when it comes to dehacked, to make things easier for wad authors - as long as it is reasonable to do so.

Here is the specification in EE from 2003:
```
A new field has been added to mobjinfo_t, bloodcolor, and a corre-
sponding mnemonic, "Blood color", added to the DeHackEd Things block.

The currently supported blood colors are as follows:

 0 - Red (normal)
 1 - Grey
 2 - Green
 3 - Blue
 4 - Yellow
 5 - Black
 6 - Purple
 7 - White
 8 - Orange
```

I think it would be best to use the same values, and fall back to red if it's not supported.

If color is an element of mobjinfo, there shouldn't be entries in the mobj type enum. A separate enumerator feels more natural, with just colors.

JadingTsunami:
> * Should we initialize to MT_BLOOD for all other monsters or rather
>   MT_NULL? This would leave the possibility to apply a special third
>   colorization with a crafted RED color translation table.

@kraflab suggestion makes sense to me as well.

> * Should we unapply the blood assignment if we detect that a monster
>   has been dehacked?

Yes, I would think this is the most natural approach. Or put explicitly: if a monster has been "DeHackEd" it would revert its blood color to the default (red) unless otherwise specified by DeHackEd.

fabiangreffrath:
Oh, I tried to be compatible to Doom Retro [1] which in turn doesn't even seem to try to be compatible to anything else. 

On the other hand, why is EE's list of colors inconsistent with Boom's array of color translation tables [2]?

[1] https://www.doomworld.com/forum/topic/66518-doom-retro-v408-updated-april-16-2021/?do=findComment&comment=2299454
[2] https://github.com/coelckers/prboom-plus/blob/e7522045b07487707eb9be8e18e18ab131eb9037/prboom2/src/v_video.c#L111

kraflab:
Given that the table there was designed for text color translation, I assume there was no reason to preserve that ordering. There's no tan blood and there's no black text. That's not to say there can't be, but it's a different use case. If you check EE you'll see it still has the table you reference there - it's just separate from the blood one.

Personally I'd go with the EE version because it's older but it doesn't mean you can't support both.

kraflab:
Incidentally, mbf21 is going to bring EE and dsda-doom (and pr+ assuming you want it) much closer, so I think the more commonalities between the two ports with respect to dehacked the better.

fabiangreffrath:
Hm, I can't seem to accomplish this without adding another pointer to `mobj_t` which would then break a dozen other assumptions, e.g. savegames.

fabiangreffrath:
Maybe you guys can help me organize my thoughts: Whenever blood is spawned, I need to transfer the bleeder's `mobjinfo[].bloodcolor` value to the blood's `mobj`, but without introducing a new element in `mobj_t` which would break e.g. savegames. I will then need an additional element in `vissprite_t` to store the bloodcolor from the blood's `mobj`.

Do you think I  could misuse the blood's `mobj.intflags` element to store the numerical blood color value?

kraflab:
I thought we were using mobj translation flags / the colored blood flag to indicate this on the blood itself.

fabiangreffrath:
Yes, but with the two `MF_TRANSLATION` flags that we are reusing we can only store up to four colors - not the entire ten that the EE spec has.

JadingTsunami:
I would agree to avoid adding another `mobj_t` pointer.

It seems to me there are really two candidate options at that point:

1. Spawn a different `mobj_t` for other blood colors.
2. Use `intflags` for extended translations (other colors).

(1) is probably a non-starter, and so my inclination is to use `intflags` anyway as it is presently underutilized and doesn't seem to have much purpose outside things like this.

kraflab:
You could take over this data: https://github.com/coelckers/prboom-plus/blob/master/prboom2/src/p_mobj.h#L375

As far as I know it's unused and then save files won't break.

fabiangreffrath:
Obviously, this is still up for discussion. 😉

kraflab:
We can rename pad to bloodcolor I think, since it isn't referenced anywhere else. The pad field doesn't exist in dsda-doom, so no clash there. Hopefully there aren't any forks that are coopting that field, but if there are then using a different name here will make an accidental clash less likely. I also think it's completely fine to have only approximate handling for the blood colors we don't have something exact for - most likely the main colors for vanilla enemies are the most common, and partial support is quite satisfactory for the scope of the port.

fabiangreffrath:
Thank you for your feedback! I think I will rename the "pad" field to "bloodcolor" then. I agree it should be alright to only approximate the more "exotic" color translations. On the other hand, it shouldn't be too much of an effort for me to add translation tables for black (i.e. dark gray), white (i.e. bright gray) and purple. What du you think?

kraflab:
If it's not much effort, then it sounds good to me

fabiangreffrath:
> Yes, I would think this is the most natural approach. Or put explicitly: if a monster has been "DeHackEd" it would revert its blood color to the default (red) unless otherwise specified by DeHackEd.

Bonus points for answering this question: How do we determine if a "monster has been dehacked"? If its flags have been changed? If one of its states/frames has been changed? if its sprites have been changed or replaced? ...

kraflab:
It's a fun question :^)

If the sprites are the same, then I'd expect the blood not to change. Most likely you just need to check for sprite replacements.

You could have more complicated things, like caco states that redirect into imp states. Then the "fake" imps will have the wrong blood even though the caco sprites weren't edited, unless you also check for changes in the caco state flows. The reverse is also true, enemies in other slots that turn into cacos, and their blood will be wrong. 👨‍🔬

fabiangreffrath:
I think we can even get away entirely without the `MF_COLOREDBLOOD` flag, if we say that `bloodcolor == 0` means "no color transition at all" instead of "color transition to red". Then we would only have to check for `bloodcolor > 0`.

fabiangreffrath:
> if there is a Thing Block in DeHackEd, the Thing has been "DeHackEd." I don't think a more elaborate solution is needed.

> If the sprites are the same, then I'd expect the blood not to change. Most likely you just need to check for sprite replacements.

> You could have more complicated things, like caco states that redirect into imp states. Then the "fake" imps will have the wrong blood even though the caco sprites weren't edited, unless you also check for changes in the caco state flows. The reverse is also true, enemies in other slots that turn into cacos, and their blood will be wrong. 👨‍🔬

I'd like to postpone this for later. I mean, we now provide a means to explicitly set blood color through DEH, so the same means could get used to reset it back to red. Guessing what's the right *implicit* thing to do is a bit more complicated than it seems, as we have seen.

Good for now?

JadingTsunami:
> I'd like to postpone this for later. I mean, we now provide a means to explicitly set blood color through DEH, so the same means could get used to reset it back to red.

Seems reasonable to me. *If* there is common feedback across multiple users that a certain method would be preferable we can incorporate it then.

PR #304, opened 2021-05-10
implement custom atexit() sequence, ported over from Chocolate Doom

fabiangreffrath:


prboom2/src/SDL/i_network.c
```diff
@@ -88,7 +91,11 @@ void I_ShutdownNetwork(void)
 void I_InitNetwork(void)
 {
   SDLNet_Init();
+#ifndef PRBOOM_SERVER
```
kraflab, line 94: Why does that need to be excluded?

fabiangreffrath: `prboom-plus-game-server` does not get linked with `SDL/i_main.c` and thus has no access to `I_AtExit()` - and wouldn't use it anyway, because is doesn't call `I_SafeExit()`.

kraflab:
Hm, this has broken the program flow in dsda-doom. `I_Quit` is never reached for some cases, like when playing a demo 🤔 
Probably because we don't call `I_SafeExit` consistently.

kraflab:
https://github.com/kraflab/dsda-doom/commit/248c15bb5760abf8d6a97fac60bc66240ee3b2a7

I'm not sure if / which other calls might need to be replaced, but this fixes the demo playback exit at least.

fabiangreffrath:
Oops! I didn't even know there were still stray plain `exit()` calls around.

kraflab:
Hm, `I_Error` isn't exiting the program for me... 🤔 
It does call I_SafeExit though... 🤔

kraflab:
Ah I see the issue. If you call `I_Error` inside code that is hit in the exit sequence, then `I_SafeExit` says "oh, I already exited, so I am a no op". That means `I_Error` returns, and doesn't yield an exit. So wherever it was called from proceeds like there wasn't an error.

fabiangreffrath:
I guess I need to set `has_exited` to a non-zero value earlier in `I_SafeExit()`, i.e. before any other function in the exit sequence is run that might call `I_SafeExit()` itself. The additional checks for `has_exited` in `I_Quit()` should be obsolete then and that function should be called also on error. The only thing that should probably not be called on error is `M_SaveDefaults()`, but this would mean that any settings which are changed during demo recording/playback won't get saved as well (I believe this is what e.g. Choco does.)

PR #306, opened 2021-05-11
fix switching weapons when reaching zero ammo with boom_autoswitch == 0

### LINKED ISSUE #303: A_Refire does not switch weapons at all when ammo is zero with boom_autoswitch set to 0
To reproduce, hold fire with the pistol until ammo is depleted -- it stays in its WeaponReady state without bobbing until you let go of the fire button.

fabiangreffrath:
This attempts to restore original Boom behavior when `boom_autoswitch`
is disabled and checks for `done_autoswitch` only if it is enabled.

Fixes #303

### DIFF
--- a/prboom2/src/g_game.c
+++ b/prboom2/src/g_game.c
@@ -579,7 +579,7 @@ void G_BuildTiccmd(ticcmd_t* cmd)
 
   // Make Boom insert only a single weapon change command on autoswitch.
   if ((!demo_compatibility && players[consoleplayer].attackdown && // killough
-       !P_CheckAmmo(&players[consoleplayer]) && !done_autoswitch && boom_autoswitch) ||
+       !P_CheckAmmo(&players[consoleplayer]) && (!boom_autoswitch || !done_autoswitch)) ||
        gamekeydown[key_weapontoggle])
   {
     newweapon = P_SwitchWeapon(&players[consoleplayer]);           // phares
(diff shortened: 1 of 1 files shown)


fabiangreffrath:
I need a second pair of eyes on this, please!

Please confirm that after this change the code behaves like the original code from Pr+ 2.5.1.7um if `boom_autoswitch == 0`:

https://github.com/coelckers/prboom-plus/blob/6c07d7924678eefe739d498d8d8b771deaf97c7c/prboom2/src/g_game.c#L566-L567

and that it behaves like the code from the original [pull request](https://github.com/coelckers/prboom-plus/pull/61) if `boom_autoswitch == 1`:

https://github.com/coelckers/prboom-plus/blob/2c59b69b2b1bed3d59f8fadda2dea96ee26c666d/prboom2/src/g_game.c#L571-L573

JadingTsunami:
It appears to work; the logic breaks down to:

```
2.5.1.7um: Always '1' (condition ignored in a logical AND chain because it wasn't there)
original PR: !done_autoswitch && boom_autoswitch
this PR: (!boom_autoswitch || !done_autoswitch)

b = boom_autoswitch
d = done_autoswitch
2 = 2.5.1.7um
o = original PR
t = this PR

bd   2 o t
00 | 1 0 1
01 | 1 0 1
10 | 1 1 1
11 | 1 0 0 
```

I think it may be easier to understand why this is the case from a logic standpoint when written as `!(boom_autoswitch && done_autoswitch)` as the only case this code tries to prevent is the case where *both* those elements are true. This is especially confusing since the original author wrote it (incorrectly) as taking the code path only when a *different* condition was true that looks like the author wanted the code path taken only when `done_autoswitch` wasn't true.

Or in other words, "don't switch weapons if we're in autoswitch mode and we already finished switching," which is what I understood the original PR was trying to fix.

PR #313, opened 2021-05-22
Fix levelstat crash

kraflab:
The levelstat code assumes that you can only reach 32 maps in one session - on your 33rd exit, you will start touching memory outside the allocated array.


fabiangreffrath:
Thanks!

PR #315, opened 2021-05-24
add autoload per PWAD feature

rfomin:
Add autoload from subdirectories with PWAD names. For example:
```
autoload/
├── aaliens.wad/
│   ├── aaliens_ws.wad
│   ├── umapinfo.lmp

```

### DIFF
--- a/prboom2/src/d_main.c
+++ b/prboom2/src/d_main.c
@@ -1377,7 +1377,7 @@ static void L_SetupConsoleMasks(void) {
 
 static char *autoload_path = NULL;
 
-static char *GetAutoloadDir(const char *iwadname)
+static char *GetAutoloadDir(const char *iwadname, dboolean createdir)
 {
     char *result;
     int len;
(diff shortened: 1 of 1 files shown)


prboom2/src/d_main.c
```diff
@@ -1400,19 +1400,35 @@ static char *GetAutoloadDir(const char *iwadname)
...
     mkdir(result);
 #else
     mkdir(result, 0755);
 #endif
+    }
 
     return result;
 }
 
+const char *BaseName(const char *filename)
```
fabiangreffrath, line 1415: Static, please.

fabiangreffrath (changes requested):

fabiangreffrath:
Awesome, thank you! 

What's needed now is the ability to add a system-wide autoload directory on non-WIN32 systems, e.g. under /usr/share/prboom-plus/prboom-plus (this applies to Woof! as well, btw).

Then we could jnstall the UMAPINFO lump files for nerve.wad and masterlevels.wad there and finally remove all the hard-coded special treatment for NRFTL from the sources.

rfomin:
> What's needed now is the ability to add a system-wide autoload directory on non-WIN32 systems, e.g. under /usr/share/prboom-plus/prboom-plus (this applies to Woof! as well, btw).

I thought this does it:
https://github.com/fabiangreffrath/woof/blob/49253820e3e28c8684194317e48b7b6d40874ef1/CMakeLists.txt#L255
I don't have non-WIN32 system, except maybe MSYS.

fabiangreffrath:
But we currently only search for autoload subdirectories in the same directory where the config is saved.

rfomin:
I've never run Doom outside of DOS/Win32, can't help here, sorry.

kraflab:
This feature doesn't work with the launcher, as far as I can tell. The launcher has been a serious thorn in my side with heretic and it's showing up here as well.

Basically the launcher does some unexpected behaviour. Here's a possible scenario:

You have doom2, plutonia, and a pwad. When pr+ starts with no options, it picks doom2 because it's the first automatic match - _it loads this as the iwad_. There is no pwad. Then we hit the autoload code which autoloads the wad files for doom2. _Then_ the launcher runs. The user now selects plutonia and the pwad, which resets the wad file lists (removing the autoloads already done). Now the iwad is plutonia and there is a pwad. When you run the deh autoloading, you autoload for plutonia and the pwad.

PR #317, opened 2021-05-26
fix autoload per PWAD with launcher

rfomin:
Fixes the issue mentioned here: https://github.com/coelckers/prboom-plus/pull/315#issuecomment-848245995

### DIFF
--- a/prboom2/src/d_main.c
+++ b/prboom2/src/d_main.c
@@ -1466,7 +1466,7 @@ static void AutoLoadWADs(const char *path)
 
 // auto-loading of .wad files.
 
-static void D_AutoloadIWadDir()
+void D_AutoloadIWadDir()
 {
   char *autoload_dir;
 
--- a/prboom2/src/d_main.c
+++ b/prboom2/src/d_main.c
@@ -1458,7 +1458,7 @@ static void AutoLoadWADs(const char *path)
         {
             break;
         }
-        D_AddFile(filename,source_pwad);
+        D_AddFile(filename,source_auto_load);
     }
 
     I_EndGlob(glob);
--- a/prboom2/src/d_main.c
+++ b/prboom2/src/d_main.c
@@ -1483,17 +1483,15 @@ static void D_AutoloadIWadDir()
 
 static void D_AutoloadPWadDir()
 {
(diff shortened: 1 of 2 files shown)


prboom2/src/e6y_launcher.c
```diff
@@ -655,10 +655,12 @@ static dboolean L_PrepareToLaunch(void)
     index = (int)SendMessage(launcher.listIWAD, CB_GETITEMDATA, index, 0);
     if (index != CB_ERR)
     {
+      extern void D_AutoloadIWadDir();
       char *iwadname = PathFindFileName(launcher.files[index].name);
       history = malloc(strlen(iwadname) + 8);
       strcpy(history, iwadname);
       AddIWAD(launcher.files[index].name);
+      D_AutoloadIWadDir();
```
fabiangreffrath, line 663: Is this now called twice, i.e. once here from `LauncherShow()` and once earlier in `D_DoomMainSetup()`?

prboom2/src/d_main.c
```diff
@@ -1483,17 +1483,15 @@ static void D_AutoloadIWadDir()
 
 static void D_AutoloadPWadDir()
 {
-  int p = M_CheckParm("-file");
-  if (p)
-  {
-    while (++p != myargc && myargv[p][0] != '-')
+  int i;
+  for (i = 0; i < numwadfiles; ++i)
+    if (wadfiles[i].src == source_pwad)
```
fabiangreffrath, line 1488: We should probably change the second argument to `D_AddFile()` in `AutoLoadWADs()` to `source_auto_load`, because else I am afraid we end up auto-loading WADs for auto-loaded WADs.

fabiangreffrath (changes requested):

prboom2/src/e6y_launcher.c
fabiangreffrath: Alright, so this takes account of the `wadfiles[]` array being killed and then rebuilt by the launcher, right?

rfomin: Yes. I tested it on Windows, seems to work.

fabiangreffrath:
Thank you!

PR #330, opened 2021-06-09
Sound Quirks Fix

Terraformer9x:
Changes an `or` statement to an `and` in a check for stopping sound. The 'kill old sound' code is supposed to emulate a quirk with item pickup sounds however it was applying the quirk to all sounds. Changing the 'or' to an 'and' fixes this. I know that the sound quirks was changed to be a setting separated from -complevel but I think it'd still be a nice addition to have this fix for the quirks.

### DIFF
--- a/prboom2/src/s_sound.c
+++ b/prboom2/src/s_sound.c
@@ -326,7 +326,7 @@ static void S_StartSoundAtVolume(degenmobj_t *origin, int sfx_id, int volume)
   // kill old sound
   for (cnum=0 ; cnum<numChannels ; cnum++)
     if (channels[cnum].sfxinfo && channels[cnum].origin == origin &&
-        (default_comp[comp_sound] || channels[cnum].is_pickup == is_pickup))
+        (default_comp[comp_sound] && channels[cnum].is_pickup == is_pickup))
       {
         S_StopChannel(cnum);
         break;


fabiangreffrath:
Hm, while this looks plausible, do you have any evidence that changing the logic from `or` to `and` is fixing the quirk as it is supposed to be (as you name it)?

fabiangreffrath:
NB: This is where the feature has been introduced:
https://github.com/coelckers/prboom-plus/commit/38dd4577df0314553c785ddcfd847bcb4770c52f

Terraformer9x:
I can show video of the change. DSDA-Doom is used here but the change is practically the same.

I made a small map with a few zombiemen to since this bug is mostly prevalent when fighting zombies. Complevel 2 was used which has comp_sound enabled by default.

With `or`: https://cdn.discordapp.com/attachments/410925031602782209/852101380390125568/before.mp4
With `and`: https://cdn.discordapp.com/attachments/410925031602782209/852101391236464660/after.mp4

kraflab:
https://github.com/chocolate-doom/chocolate-doom/blob/master/src/doom/s_sound.c#L506

Aren't we emulating this line from vanilla, which indeed kills all sounds?

fabiangreffrath:
> Aren't we emulating this line from vanilla, which indeed kills all sounds?

Yep, that was my concern!

fabiangreffrath:
@Terraformer9x Could you record a similar video with DOOM.EXE running in Dosbox?

Terraformer9x:
Sure, here it is under DOSBox: https://cdn.discordapp.com/attachments/410925031602782209/852148094689607680/dos.mp4

kraflab:
Can you make a demo and play it back in dos and prboom+ so we can see the difference for an exact comparison?

kraflab:
(and if you use a custom wad, the wad and demo too so I can check it)

kraflab:
Just from looking at the code, this PR appears to both break the old behaviour and break the new behaviour, because the sound coverage doesn't match the vanilla code and it also removes the special pickup handling added afterwards (since the and means nothing happens when comp_sound is off).

Terraformer9x:
Here's the wad and a demo: https://www.dropbox.com/s/rqec38jdvjiomtj/SFXTest.zip?dl=0

Hmm, alright. I'll say this is PR a no-go then if that's the case. There's definitely something with that code though that's implemented weirdly to where it's affecting more than just sounds related to item pickup with comp_sound.

kraflab:
According to the code, the issue should affect all sounds in vanilla, and only pickups in boom+ (i.e., the pickup-exclusive logic is unrelated to comp_sound). That's why it's an `||`. But I'm not dismissing the possibility that there is something wrong - I just don't see where the error is looking at the code itself.

fabiangreffrath:
I agree that the sfx in the "before" video sound more similar to the "dos" version, though code-wise this is still not the right fix. 🤔

fabiangreffrath:
Is `S_getChannel()` probably mising a similar `default_comp[comp_sound] ||` statement to its `channels[cnum].is_pickup == is_pickup` check?

fabiangreffrath:
> Is `S_getChannel()` probably missing a similar `default_comp[comp_sound] ||` statement to its `channels[cnum].is_pickup == is_pickup` check?

```patch
--- a/prboom2/src/s_sound.c
+++ b/prboom2/src/s_sound.c
@@ -805,7 +805,7 @@ static int S_getChannel(void *origin, sfxinfo_t *sfxinfo, int is_pickup)
   // Find an open channel
   for (cnum=0; cnum<numChannels && channels[cnum].sfxinfo; cnum++)
     if (origin && channels[cnum].origin == origin &&
-        channels[cnum].is_pickup == is_pickup)
+        (default_comp[comp_sound] || channels[cnum].is_pickup == is_pickup))
       {
         S_StopChannel(cnum);
         break;
```

kraflab:
Aha, you may be right there 👀

fabiangreffrath:
@Terraformer9x Could you please test the patch I posted yesterday (instead of your own approach)? Thanks!

Terraformer9x:
Yes, I've tested it now but it didn't work. However, while testing, I've found out what's exactly causing the sound cut off. The pickup item sound is stopping any current sound being emitted by the player. This applies to weapon noises and damage sounds.

fabiangreffrath:
Yes, that's pretty much what it says in the code.

kraflab:
My assumption is that it was intentional to make it easier to tell when you picked something up (like a key for instance)? But I wasn't there 😄

fabiangreffrath:
> Yes, I've tested it now but it didn't work. 

Honestly, I doubt that it didn't bring any improvement?

kraflab:
As we're gearing up for a final release and the port is only in maintenance mode as it is, going to close this one. Whether or not it's a bug is up for debate, but let's not change sound behaviour at the end.

PR #332, opened 2021-06-09
introduce a supplemental data directory "PRBOOMDATADIR"

fabiangreffrath:
Rationale:

 - Currently, we install `prboom-plus.wad` into `DOOMWADDIR` which
   defaults to `/usr/share/games/doom`. However, this is the exact
   directory where all other engines also expect to find their WAD
   files. Since `prboom-plus.wad` is very specific to PrBoom+ and of very
   little use for other engines, I feel it is safer to keep this file in
   a separate directory that belongs to PrBoom+ only.

   Currently, this defaults to `/usr/share/prboom-plus`.

 - I am going to add this directory as a base search path for global,
   i.e. system-wide installed, autoload data.

   As a very specific example, I am going to remove all hard-coded
   special-casing for NERVE.wad and replace this with an UMAPINFO.lmp
   file that gets autoloaded whenever the corresponding PWAD is loaded
   from the command line (except, of course, if the `-noload`
   parameter was also given - which needs to be fixed first).

### DIFF
--- a/prboom2/CMakeLists.txt
+++ b/prboom2/CMakeLists.txt
@@ -179,6 +179,7 @@ set(CMAKE_REQUIRED_INCLUDES ${CMAKE_REQUIRED_INCLUDES_PREV})
 set(CMAKE_REQUIRED_LIBRARIES ${CMAKE_REQUIRED_LIBRARIES_PREV})
 
 set(DOOMWADDIR "${CMAKE_INSTALL_PREFIX}/share/games/doom" CACHE PATH "Path to look for WAD files")
+set(PRBOOMDATADIR "${CMAKE_INSTALL_PREFIX}/share/${PACKAGE_TARNAME}" CACHE PATH "Path to install supplemental files")
 
 option(SIMPLECHECKS "Enable checks which only impose significant overhead if a posible error is detected" ON)
 option(ZONEIDCHECK "Enable id checks on zone blocks, to detect corrupted and illegally freed blocks" ON)
--- a/prboom2/cmake/config.h.cin
+++ b/prboom2/cmake/config.h.cin
@@ -5,6 +5,7 @@
 #cmakedefine PACKAGE_STRING "@PACKAGE_STRING@"
 
 #cmakedefine DOOMWADDIR "@DOOMWADDIR@"
+#cmakedefine PRBOOMDATADIR "@PRBOOMDATADIR@"
 
 #cmakedefine WORDS_BIGENDIAN
 
--- a/prboom2/data/CMakeLists.txt
+++ b/prboom2/data/CMakeLists.txt
@@ -367,4 +367,4 @@ add_custom_command(
     DEPENDS rdatawad ${WAD_SRC}
 )
 add_custom_target(prboomwad DEPENDS ${WAD_DATA_PATH})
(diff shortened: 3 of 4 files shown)


kraflab:
If we start autoloading umapinfo lumps, that will force demos into umapinfo format - even though they are not umapinfo wads. I don't think this is something a user would expect. That's unless there are some guard clauses that avoid that behaviour for the cases you mention (in which case you'd end up just moving the hard-coded handling from one place to another?)

fabiangreffrath:
Currently, all demo-critical special-casing for NERVE.wad is guarded with a `singleplayer` check. This means that players are forced to record Vanilla demos (most notably affecting level transition) with this WAD - whether they intend to or not. This would still be the case with the approach I am suggesting if the `-noload` parameter is given. I get your point about recording UMAPINFO demos by default, i.e. if this parameter isn't explicitly given, but I am currently not sure how to mitigate this.

> in which case you'd end up just moving the hard-coded handling from one place to another

I don't think that collecting all WAD-specific deviations from Vanilla in a single text file can be called hard-coding, but this is probably not what you meant?

kraflab:
I meant that you'd need to do that if you want to *not* record umapinfo demos by default for those wads. Also, won't using umapinfo instead of the hard-coded switches break demo compatibility for non-umapinfo demos for nerve? Umapinfo won't be loaded so any logical decision making in the code would need to still be there

fabiangreffrath:
Nope, as I mentioned earlier, any game-breaking desicion regarding NERVE is guarded by a `singleplayer` check which excludes demo recording and playback as well as network games. So, technically, all NERVE demos recorded with current Pr+ will still work with my proposed changes if `-noload` is given.

kraflab:
It doesn't feel right to me to introduce a change that requires a parameter to play back old demos, especially for an official wad that has lots of demos. But I think I will need to see it in action. That's a separate topic from this PR anyway though.

kraflab:
There's definitely an inherent benefit to potentially generalizing everything with built in umapinfo lumps so that even code that diverges between doom2 and doom can be reduced, it's just a question of how to do that without disrupting anything. Another option is to autodetect that a umapinfo lump has no impact on compatibility and thus avoid using its format for demos, and load it anyway for old demos.

JadingTsunami:
First, I like the idea of this PR. And having UMAPINFO lumps to resolve general compatibility issues seems like the right long-term direction to me. As far as defaulting recording demos in the "new" format, I don't see a good way around it. At some point we have to cross that bridge.

> I am currently not sure how to mitigate this.

Is there a disadvantage to invoking the original behavior as a fallback when a demo without the new header is played back?

fabiangreffrath:
I probably shouldn't have revealed my long-term goals in the commit message to this PR, as it is still strictly unrelated at this point. This is merely step zero. We need a directory to put our own engine-specific data files without interfering with other engines or their data files.

JadingTsunami:
> I probably shouldn't have revealed my long-term goals in the commit message to this PR, as it is still strictly unrelated at this point. This is merely step zero. We need a directory to put our own engine-specific data files without interfering with other engines or their data files.

Yes, this is a very good point. Since it is already broken up that way I don't see any reason not to move forward with this. We can cross other bridges later.

PR #333, opened 2021-06-10
prepare for different autoload base directories

fabiangreffrath:
This may be considered the second part of #332, although both parts
are not connected yet.

I'd like to prepare for iterating through a number of base directories
for autoloading files, namely PRBOOMDATADIR first and then the user
config directory second. Currently, this only returns the autoload
directory based on the user config directory - which corresponds to
what we currently have.

Even if we decide against autoloading based on engine-wide decision,
instead of only user decision, this still has the immediate benefit of
properly respecting the `-noload` parameter.

### DIFF
--- a/prboom2/src/d_main.c
+++ b/prboom2/src/d_main.c
@@ -1385,12 +1385,13 @@ static void L_SetupConsoleMasks(void) {
 // Calculate the path to the directory for autoloaded WADs/DEHs.
 // Creates the directory as necessary.
 
-static char *autoload_path = NULL;
-
-static char *GetAutoloadDir(const char *iwadname, dboolean createdir)
+static char *GetAutoloadBaseDir(unsigned int iter)
 {
-    char *result;
+    static char *autoload_path = NULL;
     int len;
+
+    if (M_CheckParm("-noload"))
+      return NULL;
 
     if (autoload_path == NULL)
     {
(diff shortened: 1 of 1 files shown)


fabiangreffrath:
This would be the next step, i.e. iterate through a list of directories (currently only two) as bases for the autoload directories:
```patch
diff --git a/prboom2/src/d_main.c b/prboom2/src/d_main.c
index a43afefa..d528cd43 100644
--- a/prboom2/src/d_main.c
+++ b/prboom2/src/d_main.c
@@ -1387,35 +1387,51 @@ static void L_SetupConsoleMasks(void) {
 
 static char *GetAutoloadBaseDir(unsigned int iter)
 {
-    static char *autoload_path = NULL;
-    int len;
+    static char **autoload_paths = NULL;
+    static const struct {
+        const char *dir;
+        const char *(*func)(void);
+    } basedirs[] = {
+        {PRBOOMDATADIR}, // supplemental data directory
+        {NULL, I_DoomExeDir}, // config directory
+    };
+    static const size_t num_basedirs = sizeof(basedirs)/sizeof(*basedirs);
 
     if (M_CheckParm("-noload"))
       return NULL;
 
-    if (autoload_path == NULL)
+    if (autoload_paths == NULL)
     {
-        const char* exedir = I_DoomExeDir();
-        len = doom_snprintf(NULL, 0, "%s/autoload", exedir);
-        autoload_path = malloc(len+1);
-        doom_snprintf(autoload_path, len+1, "%s/autoload", exedir);
+        int i;
+
+        autoload_paths = malloc(num_basedirs * sizeof(*autoload_paths));
+
+        for (i = 0; i < num_basedirs; i++)
+        {
+            int len;
+            const char* dir;
+
+            if (basedirs[i].func)
+                dir = basedirs[i].func();
+            else
+                dir = basedirs[i].dir;
+
+            len = doom_snprintf(NULL, 0, "%s/autoload", dir);
+            autoload_paths[i] = malloc(len+1);
+            doom_snprintf(autoload_paths[i], len+1, "%s/autoload", dir);
+        }
     }
 
 #ifdef _WIN32
-    mkdir(autoload_path);
+    mkdir(autoload_paths[iter]);
 #else
-    mkdir(autoload_path, 0755);
+    mkdir(autoload_paths[iter], 0755);
 #endif
 
-    switch (iter)
-    {
-        case 0:
-            return autoload_path;
-            break;
-        default:
-            return NULL;
-            break;
-    }
+    if (iter < num_basedirs)
+        return autoload_paths[iter];
+    else
+        return NULL;
 }
 
 static char *GetAutoloadDir(const char *base, const char *iwadname, dboolean createdir)
```

PR #335, opened 2021-06-15
fix playback of complevel 11 demo with UMAPINFO

rfomin:
`header_p` used to determine complevel from signature here: https://github.com/coelckers/prboom-plus/blob/336a7ee69fa1e0e63a611fd3a6259ca56f3bfdc6/prboom2/src/g_game.c#L4062

How to reproduce: record complevel 11 demo with loaded UMAPINFO and play it back.


kraflab:
Good catch

PR #336, opened 2021-06-15
fix play UMAPINFO demo without UMAPINFO loaded

rfomin:
Not sure if this a bug.


kraflab:
Is there any context where we might read the header of a demo before loading the wads? E.g., in the windows launcher - it does read a demo supplied to it before loading wads, but I don't know if it runs through the header code.

rfomin:
> Is there any context where we might read the header of a demo before loading the wads? E.g., in the windows launcher - it does read a demo supplied to it before loading wads, but I don't know if it runs through the header code.

Yes, that why `!(params & RDH_SKIP_HEADER)` check is here.

kraflab:
aha ok 😄

PR #340, opened 2021-06-22
UMAPINFO: fix default intermission

rfomin:
If, for example, MAP06 has UMAPINFO entry, there is currently no way to keep the default intermission screen.

Test wad: [intermission.zip](https://github.com/coelckers/prboom-plus/files/6692451/intermission.zip)

After this change, this is no longer the case:
https://github.com/coelckers/prboom-plus/blob/d7e0101a63fd2174a0d10b4444b78981d3b1160e/prboom2/doc/umapinfo.txt#L159-L160

### DIFF
--- a/prboom2/doc/umapinfo.txt
+++ b/prboom2/doc/umapinfo.txt
@@ -323,7 +323,7 @@ Thingtypes:
 Revisions:
 
 Rev 2.1 (@rfomin, June 22 2021)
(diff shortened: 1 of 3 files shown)


fabiangreffrath (approved):

fabiangreffrath:
If this changes the spec, please update it accordingly.

fabiangreffrath (changes requested):

fabiangreffrath (approved):

Issue #380, opened 2021-08-17
Sky floor "bleeding" in GL mode


XaserAcheron:
Sky rendering has been improved a ton in recent prboom builds (big props!), but there's still one sticky issue that persists. For better or worse, in Doom's software renderer, the good ol' "sky hack" only applies to ceilings, not floors (i.e. if you have two adjacent F_SKY1 sectors with different floor heights, they will _not_ be merged into a single plane) -- in prboom's GL mode, however, the sky hack works on the floor, which results in some textures bleeding through the sky in certain vanilla maps.

Random example from [UDTWID](https://www.doomworld.com/idgames/levels/doom/s-u/udtwid) E4M8's exit room -- here's what it looks like in Software (as intended):
![doom03](https://user-images.githubusercontent.com/4601589/129654630-1cff225f-bc32-4200-9d85-b2d900617fe2.png)

And here's GL (whoops! texture bleeds):
![doom02](https://user-images.githubusercontent.com/4601589/129654617-be67b09d-8119-4db3-9d8a-97c9733e39c6.png)

For ease of testing, here's a quick isolated testcase wad -- ideally you should see nothing but sky, but in GL the SKSPINE2 textures bleed through:
[skybleed.zip](https://github.com/coelckers/prboom-plus/files/6996620/skybleed.zip)

Shame, since I've always wished you could skyhack the floor, but... better be an opt-in feature somewhere rather than have maps go bork. :P

fabiangreffrath:
Who you gonna call? @JadingTsunami!

JadingTsunami:
Sure, this is a quick fix. But I think it will require some extensive testing. I have the fix working locally; I will try a few test maps to try and ensure its robustness.

JadingTsunami:
I created PR #381 to address this. The fix is working on my sky tricks test WAD and various rendering tricks WADs, but I would very much appreciate others checking it out as well. Rendering tricks tend to be finicky.

fabiangreffrath:
https://github.com/coelckers/prboom-plus/commit/79e89952c8c34437883317d43bbe74d853981e58

PR #381, opened 2021-08-17
OpenGL: Don't bleed walls through sky floors

JadingTsunami:
This is a prospective fix for #380. I would appreciate testing on it as many mapping tricks rely on strange software renderer sky behavior.

I have checked this against my test WAD that contains many/most of the weird sky tricks I've encountered, but there are always more.

### DIFF
--- a/prboom2/src/gl_main.c
+++ b/prboom2/src/gl_main.c
@@ -1918,8 +1918,7 @@ bottomtexture:
       }
       else
       {
-        if (bs->floorpic == skyflatnum &&// fs->floorpic != skyflatnum &&
-          bottomtexture == NO_TEXTURE && midtexture == NO_TEXTURE)
+        if (bottomtexture == NO_TEXTURE && midtexture == NO_TEXTURE)
         {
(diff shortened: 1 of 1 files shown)


De-Seppe:
This almost fixes the bridge effect used MAP01 of Back To Saturn X Episode 1.
2 small blocks (beneath the "lights") are still different then in software rendering.

Screenshot : https://drive.google.com/file/d/1CNEha4LiUtDmZ7Y7khRtSovNOpbav-Mm/view?usp=sharing

JadingTsunami:
> This almost fixes the bridge effect used MAP01 of Back To Saturn X Episode 1.
> 2 small blocks (beneath the "lights") are still different then in software rendering.
> 
> Screenshot : https://drive.google.com/file/d/1CNEha4LiUtDmZ7Y7khRtSovNOpbav-Mm/view?usp=sharing

I'm aware of that one. The tricks that would be needed to render it properly are both too complex and too unique to that case, and for those reasons I don't plan any further adjustments.

PR #386, opened 2021-08-29
Account for proper crosshair offsets when locked on to monsters.

CharlesLove:
When using the crosshair snapping feature, the dimensions of the crosshair aren't taken into effect. This leads to the crosshair being a bit down and to the right of where it should be. I've briefly detailed it here: https://www.doomworld.com/forum/post/2375077

Here's a fix for it.


fabiangreffrath:
Thank you!

PR #391, opened 2021-09-24
UMAPINFO: update map entries instead of replacing

rfomin:
Updates map entries instead of replacing them when parsing multiple UMPAINFO lumps.

### DIFF
--- a/prboom2/src/umapinfo.cpp
+++ b/prboom2/src/umapinfo.cpp
@@ -343,6 +343,86 @@ void ReplaceString(char **pptr, const char *newstring)
 	*pptr = strdup(newstring);
 }
 
+static void UpdateMapEntry(MapEntry *mape, MapEntry *newe)
+{
+	if (newe->levelname)
+	{
+		ReplaceString(&mape->levelname, newe->levelname);
+	}
+	if (newe->label)
+	{
+		ReplaceString(&mape->label, newe->label);
+	}
+	if (newe->intertext)
+	{
+		ReplaceString(&mape->intertext, newe->intertext);
+	}
+	if (newe->intertextsecret)
+	{
+		ReplaceString(&mape->intertextsecret, newe->intertextsecret);
+	}
+	if (newe->levelpic[0])
+	{
+		strcpy(mape->levelpic, newe->levelpic);
+	}
+	if (newe->nextmap[0])
+	{
+		strcpy(mape->nextmap, newe->nextmap);
+	}
+	if (newe->music[0])
+	{
+		strcpy(mape->music, newe->music);
+	}
+	if (newe->skytexture[0])
+	{
+		strcpy(mape->skytexture, newe->skytexture);
+	}
+	if (newe->endpic[0])
+	{
+		strcpy(mape->endpic, newe->endpic);
+	}
+	if (newe->exitpic[0])
+	{
+		strcpy(mape->exitpic, newe->exitpic);
+	}
+	if (newe->enterpic[0])
+	{
+		strcpy(mape->enterpic, newe->enterpic);
+	}
+	if (newe->interbackdrop[0])
+	{
+		strcpy(mape->interbackdrop, newe->interbackdrop);
+	}
+	if (newe->intermusic[0])
+	{
+		strcpy(mape->intermusic, newe->intermusic);
+	}
+	if (newe->partime)
+	{
+		mape->partime = newe->partime;
+	}
+	if (newe->nointermission)
+	{
+		mape->nointermission = newe->nointermission;
+	}
+	if (newe->numbossactions)
+	{
+		mape->numbossactions = newe->numbossactions;
+		if (mape->numbossactions == -1)
+		{
+			if (mape->bossactions) free(mape->bossactions);
+			mape->bossactions = NULL;
+		}
+		else
+		{
+			mape->bossactions = (struct BossAction *)realloc(mape->bossactions,
+				sizeof(struct BossAction) * mape->numbossactions);
+			memcpy(mape->bossactions, newe->bossactions,
+				sizeof(struct BossAction) * mape->numbossactions);
+		}
+	}
+}
+
 // -----------------------------------------------
 //
 // Parses a set of string and concatenates them
@@ -699,13 +779,13 @@ int ParseUMapInfo(const unsigned char *buffer, size_t length, umapinfo_errorfunc
 			}
 		}
 
-		// Does this property already exist? If yes, replace it.
+		// Does this property already exist? If yes, update it.
 		for(i = 0; i < Maps.mapcount; i++)
 		{
 			if (!strcmp(parsed.mapname, Maps.maps[i].mapname))
 			{
-				FreeMap(&Maps.maps[i]);
-				Maps.maps[i] = parsed;
+				UpdateMapEntry(&Maps.maps[i], &parsed);
+				FreeMap(&parsed);
 				break;
 			}
 		}


fabiangreffrath:
I am all for it, just saying...

fabiangreffrath (approved):

kraflab:
This requires updating the umapinfo spec and documenting that the behaviour is changed, especially since this potentially breaks demo compatibility. And that can't be done without someone owning the umapinfo spec. I think @XaserAcheron was interested in taking that responsibility and organizing these kinds of discussions. A bug caused by how woof supports nerve is not instant justification to break umapinfo compatibility. Umapinfo is already in a state where different ports implement different pieces of it at different spec points, partially because of so many ad hoc changes dropped into prboom+.

rfomin:
> Umapinfo is already in a state where different ports implement different pieces of it at different spec points, partially because of so many ad hoc changes dropped into prboom+.

The GZDoom implementation is not updated. PrBoom+/DSDA-Doom/Woof is ok, Doom Retro implementation is incomplete. I believe there are no other ports that implement UMAPINFO.

kraflab:
Incidentally, I don't agree with this change at all. I think the replacement behaviour is more natural.

rfomin:
> the replacement behaviour is more natural.

But why? In most other cases UMAPINFO allows default values. You can write like this:
```
map MAP01
{
    levelname = "custom 1"
}
```
and everything will work with default engine values. But if do the same for PWAD with built-in UMAPINFO lump everything will reset to 0.

fabiangreffrath:
Yep, consistency is s pretty strong reason. In this context, this is how DEHACKED already works, so it might be considered pretty much expected behavior in the first place. Also, it avoids the necessity to hard-code specific behavior for NRFTL in our port which, as I understand it, is in line with the original motivation for the introduction of the UMAPINFO feature.

kraflab:
If we want to compare to other standards, it should be to other mapinfo implementations, not to dehacked. Hexen uses replacement for MAPINFO lumps. Do you know how zdoom / eternity work?

kraflab:
From EE for instance:
```
an entry for a map will be searched for starting with the newest EMAPINFO lump in the wad
directory (this will be the last such lump in the last wad to be loaded). If an entry is found
for the map in that lump, it will be used and any others are ignored.
```

kraflab:
One thing you could do for woof is to make the "default" value based on a different base file for different contexts, which I guess is how something like zdoom works. Then the "default" for nerve would be what is currently created by your automatic umapinfo loading. I've been actually thinking about doing something like that in dsda-doom...I mean, ideally there would be very little code in there that checks the game when deciding progression. It should all be converted into some internal mapinfo logic and then the rest of the code base could be much more agnostic.

rfomin:
> If we want to compare to other standards, it should be to other mapinfo implementations, not to dehacked. 

GZDoom doesn't use engine defaults in MAPINFO. I think @coelckers explicitly chooses a different approach in UMAPINFO.

> One thing you could do for woof...

I'm sure we can find some solution for Woof without UMAPINFO changes.

fabiangreffrath:
> I'm sure we can find some solution for Woof without UMAPINFO changes.

I am not going to hard-code any NRFTL-specific behavior into the engine, because this is exactly what we have UMAPUNFO support for.

rfomin:
I mean the Woof issue is not my main motivation for this change. I have been using UMAPINFO for a while and help others with this and the defaults handling always confuses people. Also, I want to eliminate these exceptions: https://github.com/coelckers/prboom-plus/blob/b3eb71424d2775d16f74190161fa6e05fb7d7973/prboom2/doc/umapinfo.txt#L144-L157

fabiangreffrath:
How about we selectively enable updating non-critical entries like e.g. map title and music but nothing affecting map progre?

XaserAcheron:
Can't comment on this in full just yet, but a driveby suggestion: add a new "cumulative = true" flag (better name suggestions welcome) that selectively enables this proposed new behavior. Better than trying to do breaking changes for a cross-port feature in a demo-compatible port.

rfomin:
I think there are no UMAPINFO demos in the DSDA archive, so we are not breaking anything yet.

kraflab:
On the contrary, there are lots of umapinfo demos on dsda now, as it has started to be adapted by some high profile mappers. It doesn't matter anyway though, because there will always be tons of people using the current versions of prboom+ and dsda-doom that never upgrade. You have to consider that the existing behaviour will always be there in parallel with any change you want to make.

rfomin:
Well, then it's probably too late for this kind of change. I thought that we still have some time.

rfomin:
> Are those issues addressed in this PR?

No, but I think should make UMAPINFO more consistent in the future.

JadingTsunami:
> Can't comment on this in full just yet, but a driveby suggestion: add a new "cumulative = true" flag (better name suggestions welcome) that selectively enables this proposed new behavior. Better than trying to do breaking changes for a cross-port feature in a demo-compatible port.

In a similar vein, embedding a UMAPINFO spec version (which I think @kraflab has suggested before) would be a general-purpose solution. There are and will be other issues of this kind. The behavior can adapt to the spec version.

XaserAcheron:
Yep, that's even better -- a top-level "version" keyword similar to ZScript that denotes the spec version would provide a well-supported path for these kinds of behavior updates.

rfomin:
> a top-level "version" keyword

So, in the case of multiple UMAPINFO lumps, if one of them has `version > 2.1` switch to update entries mode, otherwise replace entries.

kraflab:
When it comes to changing the merging behaviour, there isn't really a great solution. If you created a umapinfo patch right now, you expect it to override anything else, but that won't happen if you use it with a new wad that uses a new spec version, if you made the highest spec version win (and this behaviour will even then only happen in specific ports / versions). So then one patch has different behaviour depending on the source wad you apply it to? I'd still like some input about how other MAPINFO specs work, since it seems like the EE one does full replacement as well. It would be awkward if you need to use MAPINFO differently for each different flavour of it. But if there's already no consistency, then that's not a plus or minus.

It's a major conceptual change and isn't a bug fix or iteration (it's quite clearly spelled out how it works in the comments of the parser, so this behaviour is certainly by design, whether you agree with the design or not). To me, that kind of thing requires a group to take a look at UMAPINFO and reenvision this spec, making whatever tweaks and updates make sense as well with regards to the issues rfomin brought up, and create a v3. Ideally with some format that will fail in old umapinfo parsers so those users know they are not getting the intended behaviour from the patch. But that's not really pleasant either... 🤷 

Serious question time: is UMAPINFO a nice experiment but basically a failure? Should we spin it into something "new" entirely? It's over 4 years old but hasn't really delivered, in the sense that it is not widely adopted either by source ports or by wad authors. If it wasn't already implemented in pr+ when I started dsda-doom, I doubt I would have added it there myself, because it feels like there are better paths forward in this domain.

XaserAcheron:
Let's hold on for a bit before making any large changes like this -- we need to regroup and get a spec together, but I'm extremely busy ATM and cannot drop my current stuff to kick this off yet. Should have some time early next month.

re: adoption, UMAPINFO has basically only been considered a serious feature in the last year-ish (e.g. from the point pr+um got "canonized" and dsda-doom sprang up), because up until then it was just some experimental feature in a rogue fork. Give it some time -- there are several notable projects in the works, but they're large multi-map wads so they take a while to develop.

Altazimuth:
Unless this new behaviour is specifically opt-in (as to not break compatibility with already-released and in-progress projects) I am wholly against such a move.

rfomin:
I agree with most of the points, but I think we made this change much larger than it actually is. I'm pretty sure that if we just apply this change as is, there will be no desync in the current demos.

I think there are two possible scenarios where this change matters:

1. Someone has included a multiple UMAPINFO lumps in the PWAD (probably by mistake). Or there are duplicate map entries in the same UMAPINFO lump with differences (again, probably a mistake).
2. Woof internally uses UMAPINFO lumps. Hypothetical patches for Sigil or midipacks/addons for PWADs with UMAPINFO.

If we introduce a version field, we can easily avoid potential errors.

rfomin:
> If you created a umapinfo patch right now, you expect it to override anything else, but that won't happen if you use it with a new wad that uses a new spec version, if you made the highest spec version win (and this behaviour will even then only happen in specific ports / versions). So then one patch has different behaviour depending on the source wad you apply it to?

Both the new and the old spec override everything, but the old spec resets all fields to 0 before applying changes.

> It would be awkward if you need to use MAPINFO differently for each different flavour of it. But if there's already no consistency, then that's not a plus or minus.

There is no consistency already. And I think UMAPINFO is more convenient to use.

fabiangreffrath:
I think it's clear that this wont happen anytime soon.

PR #420, opened 2021-11-06
UMAPINFO: fix entering and enterpic shown on exit levels

### LINKED ISSUE #419: UMAPINFO: "Entering" and enterpic still shown on exit levels
Reported by a user [here](https://www.doomworld.com/forum/post/2410486).

Reproducible with this minimal UMAPINFO:
```
map map01
{
 endcast = true
 intertext = "Sample text goes here"
}
map map02
{
 enterpic = "CREDIT"
}
```

One option for a fix might be a small modification in `WI_drawShowNextLoc`:
`src/wi_stuff.c:1041`
```c
  if (!secretexit && gamemapinfo->endpic[0] && (strcmp(gamemapinfo->endpic, "-") != 0))
      return;
```

rfomin:
Fixes #419

### DIFF
--- a/prboom2/src/wi_stuff.c
+++ b/prboom2/src/wi_stuff.c
@@ -1038,6 +1038,13 @@ void WI_drawShowNextLoc(void)
   int   i;
   int   last;
 
+  if (gamemapinfo != NULL &&
+      gamemapinfo->endpic[0] &&
+      strcmp(gamemapinfo->endpic, "-") != 0)
+  {
+    return;
+  }
+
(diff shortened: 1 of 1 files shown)


fabiangreffrath:
Do we need to update the spec about it or is this considered merely a fix of the implementation to match the spec?

rfomin:
> Do we need to update the spec about it or is this considered merely a fix of the implementation to match the spec?

I think we already fixed this in Rev 1.6. We may not have tested this particular scenario.

PR #427, opened 2021-11-21
launcher: identify IWADs either by IWAD tag or by PWAD tag and matchi…

fabiangreffrath:
…ng file name


fabiangreffrath:
FYI @kraflab @El-Juancho

PR #432, opened 2021-12-02
fix scaled time interpolation

fabiangreffrath:
@rfomin @kraflab Am I doing this right?

### DIFF
--- a/prboom2/src/SDL/i_main.c
+++ b/prboom2/src/SDL/i_main.c
@@ -84,12 +84,59 @@ typedef BOOL (WINAPI *SetAffinityFunc)(HANDLE hProcess, DWORD mask);
  * cphipps - much made static
  */
 
+static int basetime = 0;
+
+static int I_GetTime_MS(void)
+{
+    int ticks = SDL_GetTicks();
+
+    if (basetime == 0)
+        basetime = ticks;
+
+    return ticks - basetime;
+}
+
+int ms_to_next_tick;
+
+int I_GetTime_RealTime (void)
+{
+  int i;
+  int t = SDL_GetTicks();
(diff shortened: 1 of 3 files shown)


rfomin:
I'm sorry, I'm busy right now. I'll take a look at this later today.

rfomin:
It works in my testing, but there is a jump of several frames when changing the speed. Perhaps these lines are relevant:
https://github.com/coelckers/prboom-plus/blob/98c3615b0639acc441b6aa8f8425de0eeec79f3f/prboom2/src/d_client.c#L352-L354

fabiangreffrath:
> It works in my testing, but there is a jump of several frames when changing the speed. Perhaps these lines are relevant:

Hm, maybe this code was necessary with the old implementation but now plays against the new one?

fabiangreffrath:
Oh, wait, apparently `I_GetTime_Scaled()` is missing a calculation for `ms_to_next_tick`. Maybe that's related.

rfomin:
> Hm, maybe this code was necessary with the old implementation but now plays against the new one?

I tried to remove it - no effect. DSDA-Doom has these lines.

fabiangreffrath:
Could you please check if this new approach has fixed it for you?

rfomin:
Just tested DSDA-Doom, it has a small jump too. So I think it's fine.

fabiangreffrath:
I still consider `ms_to_next_tick` to be the culprit. Could you please check again?

rfomin:
> I still consider `ms_to_next_tick` to be the culprit. Could you please check again?

Still the same jump. I don’t know, maybe this is how it should work.

PR #436, opened 2021-12-17
UMAPINFO: Introduce the new author field.

rfomin:
The level `author` field is supported by Eternity Engine in EMAPINFO (`creator`), GZDoom in ZMAPINFO (`author`), and Doom Retro in RMAPINFO (`author`).

`musictitle` and `musiccomposer` are taken from Doom Retro.

In demo compatible ports, the status screen for "entering" in Doom 2 is very fast, barely enough time to read the name of the level.


fabiangreffrath:
Sure, why not. I am fine with this addition to the spec.

> In demo compatible ports, the status screen for "entering" in Doom 2 is very fast, barely enough time to read the name of the level.

Is this the reason why there is no coresponding change in the code to actually print the new fields?

rfomin:
> Is this the reason why there is no coresponding change in the code to actually print the new fields?

I thought maybe people don't want these changes. Sure, this is trivial, I can add this to PrBoom+/Woof.

JadingTsunami:
> > Is this the reason why there is no coresponding change in the code to actually print the new fields?
> 
> I thought maybe people don't want these changes. Sure, this is trivial, I can add this to PrBoom+/Woof.

Hmm, there are multiple musics defined per level (for intertext/etc.).

Are these music fields purely for documentation or are they displayed in some way while music is playing back?

Does it create any concern that not all music can be attributed to an author/composer?

Just posing some upfront questions so we don't thrash on this.

rfomin:
> Hmm, there are multiple musics defined per level (for intertext/etc.).

This is only for level music.

> Are these music fields purely for documentation or are they displayed in some way while music is playing back?

They are displayed on the "entering" status screen along with CWILV graphics.

> Does it create any concern that not all music can be attributed to an author/composer?

That is the question for PWAD/UMAPINFO maker.

JadingTsunami:
> That is the question for PWAD/UMAPINFO maker.

Hmm, why do you say this? If we don't supply a field for it I don't understand how the WAD author/maker is expected to handle this.

rfomin:
Sorry, I guess I don't understand. These are optional fields so that the PWAD/UMAPINFO creators can display MIDI music authors in-game if they want.

JadingTsunami:
> Sorry, I guess I don't understand. These are optional fields so that the PWAD/UMAPINFO creators can display MIDI music authors in-game if they want.

Put more simply: Why is it important to provide a means to credit the author/composer of level music but *not* important to provide a means credit the author/composer of other musics?

rfomin:
> Put more simply: Why is it important to provide a means to credit the author/composer of level music but _not_ important to provide a means credit the author/composer of other musics?

Well, I think there is no convenient place in-game to credit the authors of other musics.

JadingTsunami:
> > Put more simply: Why is it important to provide a means to credit the author/composer of level music but _not_ important to provide a means credit the author/composer of other musics?
> 
> Well, I think there is no convenient place in-game to credit the authors of other musics.

We could just pick a corner, lower-right for instance. Then it's up to the author to either exclude the field or ensure it is visible against the graphic backdrop.

rfomin:
> > > Put more simply: Why is it important to provide a means to credit the author/composer of level music but _not_ important to provide a means credit the author/composer of other musics?
> > 
> > 
> > Well, I think there is no convenient place in-game to credit the authors of other musics.
> 
> We could just pick a corner, lower-right for instance. Then it's up to the author to either exclude the field or ensure it is visible against the graphic backdrop.

I don’t know, it’s distracting. So we need to add fields for all types of intermissions? I have seen several PWADs that indicated level and music authors graphically in CWILV graphics, but I have never seen PWAD that displayed authors of an intermission music in-game.

kraflab:
Hm, mapinfo doesn't feel like the right place to have music info. There can be multiple music tracks in one level, and mapinfo doesn't cover all the places music appears. Thete are plenty of ways to credit music. If it's a big deal to you, I would suggest standardizing a music info lump that expressly assigns credit to songs, not maps.

kraflab:
Also, why is this coming in the form of a PR and not an issue or a forum thread?

JadingTsunami:
> Hm, mapinfo doesn't feel like the right place to have music info. There can be multiple music tracks in one level, and mapinfo doesn't cover all the places music appears. Thete are plenty of ways to credit music. If it's a big deal to you, I would suggest standardizing a music info lump that expressly assigns credit to songs, not maps.

I actually like this idea better too. Especially since we can have music changers which will allow multiple tracks per map.

rfomin:
Ok, I removed `musictitle` and `musiccomposer`.

> Also, why is this coming in the form of a PR and not an issue or a forum thread?

This PR is inspired by this thread: [[doomworld]](https://www.doomworld.com/forum/topic/122892-umapinfo-appreciation-thread/?do=findComment&comment=2428639)

fraansg:
Why hasn't this PR been merged yet? It's my understanding that the author field has already been adopted as part of the UMAPINFO specification (https://github.com/kraflab/umapinfo/pull/1), but hasn't been implemented by any source port yet (from what i've seen).

kraflab:
This PR either needs to copy the official spec file exactly or otherwise replace prboom+'s copy with a link to the official spec version. After that I'd agree it can be merged.

prboom2/src/wi_stuff.c
```diff
@@ -567,6 +569,13 @@ void WI_drawEL(void)
 		// CPhipps - patch drawing updated
 		V_DrawNamePatch((320 - V_NamePatchWidth(lname)) / 2, y, FB,
 			lname, CR_DEFAULT, VPT_STRETCH);
+
+		y += (5 * V_NamePatchHeight(lname)) / 4;
+	}
+
+	if (wbs->nextmapinfo && wbs->nextmapinfo->author)
```
kraflab, line 576: The spec stipulates that the author name is only printed when the levelpic doesn't exist and we are printing the level name (the above branch). In a lot of wads, the organizer includes the author in the patch along with the title.

kraflab:
The stats screen also needs to be updated (`WI_drawLF`).

prboom2/src/wi_stuff.c
fabiangreffrath: @rfomin could you apply the change requested by @kraflab above? This would be the last remaining change I would like to merge into the prboom-plus code base before nailing is coffin.

rfomin: Done in https://github.com/coelckers/prboom-plus/pull/436/commits/1bb0a74a57d27571efefb6bf12620e55ac1c045d

PR #442, opened 2021-12-24
Do not remove active plats in v1.2 compat

### LINKED ISSUE #441: Desync on JCM2-16 E2M8 v1.2 demo
WAD: [JCM2-16](https://www.doomworld.com/idgames/levels/doom/j-l/jcm2-16)
demo: DEMO1 in JCM-E2M8.WAD in [DOOM_J2D.ZIP](https://archive.org/download/cdrom-1001-nights-doom/1001NightsOfDoom1995wickedSensations.iso/LMPS%2FDOOM_J2D.ZIP)

This HMP demo plays back correctly to the end (10:56) with v1.2 vanilla and my private Crispy fork that incorporates all of SmileTheory's v1.0-1.2 fixes (PRs [#1205](https://github.com/chocolate-doom/chocolate-doom/pull/1205) and [#1207](https://github.com/chocolate-doom/chocolate-doom/pull/1207) on the Choco repo). 

The demo desyncs under current prb+ and Crispy masters. It starts to go out visibly of sync around 4:15-4:20, when the player is supposed to enter a hallway from outside and fight the Cacos in there, and by the 4:40 mark it goes haywire. The player never makes it into the hallway.

I found that I could fix this for the Crispy master by applying ST's commit [3fe7123](https://github.com/chocolate-doom/chocolate-doom/pull/1207/commits/3fe712316d947353a1576dd0bb91d3a395d8513b) (Emulate pre-v1.25 always active plats) from PR #1207; however, this single change is not enough in prb+'s case. Apparently it's missing at least one other critical fix from the same PR, probably [eebc685](https://github.com/chocolate-doom/chocolate-doom/pull/1207/commits/eebc68589a4efc0fc9b5b4a7d8e5c077053b6ca6) (Fix inaccurate v1.0/v1.1/v1.2 sight calculations) or [046ae69](https://github.com/chocolate-doom/chocolate-doom/pull/1207/commits/046ae694c2ef47a824fb266db830f2dc353bba42) (Emulate pre-v1.25 lowerAndChange behavior).

tpoppins:
This is an adaptation of SmileTheory's commit [3fe7123](https://github.com/chocolate-doom/chocolate-doom/pull/1207/commits/3fe712316d947353a1576dd0bb91d3a395d8513b) (Emulate pre-v1.25 always active plats) from PR [#1207](https://github.com/chocolate-doom/chocolate-doom/pull/1207) on the Choco repo. Fixes #441.

### DIFF
--- a/prboom2/src/e6y.c
+++ b/prboom2/src/e6y.c
@@ -87,9 +87,7 @@
 #include "d_deh.h"
 #include "e6y.h"
 
-#ifdef _WIN32
-#include "WIN/win_fopen.h"
-#endif
+#include "m_io.h"
 
 dboolean wasWiped = false;
 
@@ -220,7 +218,7 @@ void e6y_assert(const char *format, ...)
   va_list argptr;
   va_start(argptr,format);
   //if (!f)
-    f = fopen("d:\\a.txt", "ab+");
+    f = M_fopen("d:\\a.txt", "ab+");
   vfprintf(f, format, argptr);
   fclose(f);
   va_end(argptr);
@@ -272,6 +270,7 @@ prboom_comp_t prboom_comp[PC_MAX] = {
   {0x00000000, 0x02050007, 0, "-do_not_use_misc12_frame_parameters_in_a_mushroom"},
   {0x00000000, 0x02050102, 0, "-apply_mbf_codepointers_to_any_complevel"},
   {0x00000000, 0x02050104, 0, "-reset_monsterspawner_params_after_loading"},
+  {0x02050104, 0x02060200, 0, "-remove_active_plats"},
 };
 
 void e6y_InitCommandLine(void)
@@ -967,7 +966,7 @@ void e6y_WriteStats(void)
   tmpdata_t *all;
   size_t allkills_len=0, allitems_len=0, allsecrets_len=0;
 
-  f = fopen("levelstat.txt", "wb");
+  f = M_fopen("levelstat.txt", "wb");
   
   all = malloc(sizeof(*all) * numlevels);
   memset(&max, 0, sizeof(timetable_t));
@@ -1309,12 +1308,12 @@ int GetFullPath(const char* FileName, const char* ext, char *Buffer, size_t Buff
     switch(i)
     {
     case 0:
-      getcwd(dir, sizeof(dir));
+      M_getcwd(dir, sizeof(dir));
       break;
     case 1:
-      if (!getenv("DOOMWADDIR"))
+      if (!M_getenv("DOOMWADDIR"))
         continue;
-      strcpy(dir, getenv("DOOMWADDIR"));
+      strcpy(dir, M_getenv("DOOMWADDIR"));
       break;
     case 2:
       strcpy(dir, I_DoomExeDir());
@@ -1327,49 +1326,6 @@ int GetFullPath(const char* FileName, const char* ext, char *Buffer, size_t Buff
   }
 
   return false;
-}
-#endif
-
-#ifdef _WIN32
-#include <Mmsystem.h>
-#ifndef __GNUC__
-#pragma comment( lib, "winmm.lib" )
-#endif
-int mus_extend_volume;
-void I_midiOutSetVolumes(int volume)
-{
-  // NSM changed to work on the 0-15 volume scale,
-  // and to check mus_extend_volume itself.
-  
-  MMRESULT result;
-  int calcVolume;
-  MIDIOUTCAPS capabilities;
-  unsigned int i;
-
-  if (volume > 15)
-    volume = 15;
-  if (volume < 0)
-    volume = 0;
-  calcVolume = (65535 * volume / 15);
-
-  //SDL_LockAudio(); // this function doesn't touch anything the audio callback touches
-
-  //Device loop
-  for (i = 0; i < midiOutGetNumDevs(); i++)
-  {
-    //Get device capabilities
-    result = midiOutGetDevCaps(i, &capabilities, sizeof(capabilities));
-    if (result == MMSYSERR_NOERROR)
-    {
-      //Adjust volume on this candidate
-      if ((capabilities.dwSupport & MIDICAPS_VOLUME))
-      {
-        midiOutSetVolume((HMIDIOUT)i, MAKELONG(calcVolume, calcVolume));
-      }
-    }
-  }
-
-  //SDL_UnlockAudio();
 }
 #endif
 
--- a/prboom2/src/p_plats.c
+++ b/prboom2/src/p_plats.c
@@ -115,13 +115,13 @@ void T_PlatRaise(plat_t* plat)
           {
             case blazeDWUS:
             case downWaitUpStay:
+            case raiseAndChange:
             case genLift:
               P_RemoveActivePlat(plat);     // killough
               break;
 
-            case raiseAndChange:
             case raiseToNearestAndChange:
-            // In versions <= v1.2 these platform types always remain active.
+            // In versions before v1.666 these platform types always remain active.
             if (compatibility_level > doom_12_compatibility)
             {
               P_RemoveActivePlat(plat);
--- a/prboom2/src/p_plats.c
+++ b/prboom2/src/p_plats.c
@@ -160,7 +160,6 @@ void T_PlatRaise(plat_t* plat)
             case raiseAndChange:
             case raiseToNearestAndChange:
               P_RemoveActivePlat(plat);
-
             default:
               break;
           }
--- a/prboom2/src/p_plats.c
+++ b/prboom2/src/p_plats.c
@@ -115,13 +115,18 @@ void T_PlatRaise(plat_t* plat)
           {
             case blazeDWUS:
             case downWaitUpStay:
+            case genLift:
+              P_RemoveActivePlat(plat);     // killough
+              break;
+
             case raiseAndChange:
             case raiseToNearestAndChange:
-            case genLift:
-		    // SmileTheory: in versions <= v1.2 (at least), platform types besides
-		    // downWaitUpStay always remain active.
-		    if (compatibility_level > doom_12_compatibility)
-              P_RemoveActivePlat(plat);     // killough
+            // In versions <= v1.2 these platform types always remain active.
+            if (compatibility_level > doom_12_compatibility)
+            {
+              P_RemoveActivePlat(plat);
+              break;
+            }
             default:
               break;
           }
--- a/prboom2/src/p_plats.c
+++ b/prboom2/src/p_plats.c
@@ -118,8 +118,6 @@ void T_PlatRaise(plat_t* plat)
             case raiseAndChange:
             case raiseToNearestAndChange:
             case genLift:
-		    // SmileTheory: in versions <= v1.2 (at least), platform types besides
-		    // downWaitUpStay always remain active.
 		    if (compatibility_level > doom_12_compatibility)
               P_RemoveActivePlat(plat);     // killough
             default:
@@ -159,11 +157,8 @@ void T_PlatRaise(plat_t* plat)
           {
             case raiseAndChange:
             case raiseToNearestAndChange:
-		    // SmileTheory: in versions <= v1.2 (at least), platform types besides
-		    // downWaitUpStay always remain active.
-		    if (compatibility_level > doom_12_compatibility)
-                P_RemoveActivePlat(plat);
-            break;
+              P_RemoveActivePlat(plat);
+
             default:
               break;
           }
--- a/prboom2/src/p_plats.c
+++ b/prboom2/src/p_plats.c
@@ -122,7 +122,7 @@ void T_PlatRaise(plat_t* plat)
 
             case raiseToNearestAndChange:
             // In versions before v1.666 these platform types always remain active.
-            if (compatibility_level > doom_12_compatibility)
+            if (compatibility_level > doom_12_compatibility || prboom_comp[PC_v12_REMOVE_ACTIVE_PLATS].state)
             {
               P_RemoveActivePlat(plat);
               break;
--- a/prboom2/src/p_plats.c
+++ b/prboom2/src/p_plats.c
@@ -118,6 +118,9 @@ void T_PlatRaise(plat_t* plat)
             case raiseAndChange:
             case raiseToNearestAndChange:
             case genLift:
+		    // SmileTheory: in versions <= v1.2 (at least), platform types besides
+		    // downWaitUpStay always remain active.
+		    if (compatibility_level > doom_12_compatibility)
               P_RemoveActivePlat(plat);     // killough
             default:
               break;
@@ -156,7 +159,11 @@ void T_PlatRaise(plat_t* plat)
           {
             case raiseAndChange:
             case raiseToNearestAndChange:
-              P_RemoveActivePlat(plat);
+		    // SmileTheory: in versions <= v1.2 (at least), platform types besides
+		    // downWaitUpStay always remain active.
+		    if (compatibility_level > doom_12_compatibility)
+                P_RemoveActivePlat(plat);
+            break;
             default:
               break;
           }
--- a/prboom2/src/p_plats.c
+++ b/prboom2/src/p_plats.c
@@ -118,6 +118,8 @@ void T_PlatRaise(plat_t* plat)
             case raiseAndChange:
             case raiseToNearestAndChange:
             case genLift:
+		    // SmileTheory: in versions <= v1.2 (at least), platform types besides
+		    // downWaitUpStay always remain active.
 		    if (compatibility_level > doom_12_compatibility)
               P_RemoveActivePlat(plat);     // killough
             default:
--- a/prboom2/src/e6y.h
+++ b/prboom2/src/e6y.h
@@ -255,6 +255,7 @@ enum
   PC_DO_NOT_USE_MISC12_FRAME_PARAMETERS_IN_A_MUSHROOM,
   PC_APPLY_MBF_CODEPOINTERS_TO_ANY_COMPLEVEL,
   PC_RESET_MONSTERSPAWNER_PARAMS_AFTER_LOADING,
+  PC_v12_REMOVE_ACTIVE_PLATS,
   PC_MAX
 };
 
@@ -355,10 +356,5 @@ void I_Warning(const char *message, ...);
 int I_MessageBox(const char* text, unsigned int type);
 
 dboolean SmoothEdges(unsigned char * buffer,int w, int h);
-
-#ifdef _WIN32
-extern int mus_extend_volume;
-void I_midiOutSetVolumes(int volume);
-#endif
 
 #endif
--- a/prboom2/src/p_spec.c
+++ b/prboom2/src/p_spec.c
@@ -2667,7 +2667,9 @@ void P_SpawnSpecials (void)
 
       case 17:
         // fire flickering
-        P_SpawnFireFlicker(sector);
+        // first introduced in official v1.4 beta
+        if (compatibility_level > doom_12_compatibility)
+          P_SpawnFireFlicker(sector);
         break;
     }
   }
--- a/prboom2/src/p_spec.c
+++ b/prboom2/src/p_spec.c
@@ -2667,9 +2667,7 @@ void P_SpawnSpecials (void)
 
       case 17:
         // fire flickering
-        // first introduced in official v1.4 beta
-        if (compatibility_level > doom_12_compatibility)
-          P_SpawnFireFlicker(sector);
+        P_SpawnFireFlicker(sector);
         break;
     }
   }


kraflab:
Thanks for looking into this 👍

kraflab:
Hm, I guess a concern is if there are demos recorded with the bug that will need to be addressed.

tpoppins:
I'm not aware of anyone besides myself recording prboom-plus -complevel 0 demos, and I only did four. Of those, three are unaffected by this change. The fourth (RETURN01 [E1M2 -fast](https://dsdarchive.com/files/demos/return01/19916/re12f808.zip)) is a real mongrel of a demo that never played back correctly in anything but the prb+ version it was recorded with. v1.2 emulation still had large holes at that time, not to mention that v1.2 had no support for _-fast_.

OK, Google turns up [JuanchoES' run](https://dsdarchive.com/files/demos/uac_dead/54934/uac_dead-243.zip) on UAC_DEAD and it does desync now. Heaven knows why he decided to use -cl0 when -cl2 had worked for Tatsurd, TGH and myself. Other than that I only found two more DOOM demos recorded with prb+ -cl0, both by 4shock and both play back fine. The rest is  all DSDA-Doom demos for Heretic. Hmm, there might be a concern for porting this over to DSDAD.

prboom2/src/p_plats.c
```diff
@@ -156,7 +159,11 @@ void T_PlatRaise(plat_t* plat)
           {
             case raiseAndChange:
             case raiseToNearestAndChange:
-              P_RemoveActivePlat(plat);
```
fabiangreffrath, line 159: Actually, @SmileTheory 's patch for Choco only affects the `case up` code path and the `case down` code path should already be disabled in doom_12_compatibility because of the `if (!comp[comp_floors])` guard, right?

tpoppins: You are right, I missed the comp_floors guard.
You can train a monkey to submit PRs but no amount of bananas will turn it into a programmer. ;)

prboom2/src/p_plats.c
```diff
@@ -157,6 +160,7 @@ void T_PlatRaise(plat_t* plat)
             case raiseAndChange:
             case raiseToNearestAndChange:
               P_RemoveActivePlat(plat);
+
```
fabiangreffrath, line 163: There is still a blank line left here.

prboom2/src/p_plats.c
fabiangreffrath: Thanks! 

Still, the crucial question remains: Do we want to develop complevel 0 to become more accurate for the sake of demo compatibility with Vanilla Doom 1.2? Or do we want to keep it as it is with known bugs for the sake of demo compatibility with itself?

In Woof we faced a similar question and decided to go the second route, i.e. we replaced original MBF code with the pendant from PrBoom+ in order to remain demo compatible with complevel 9.

kraflab:
```c
if (heretic && plat->type == raiseToNearestAndChange) break;
P_RemoveActivePlat(plat);     // killough
```
😄 I should have written this down on my list of "why is it different in heretic?" notes when I implemented that! Should it really happen for every case like in this PR or should it mirror this condition from heretic? :^)

prboom2/src/p_plats.c
tpoppins: Fixed in ef86eb4.

tpoppins:
Thank you for the bump, kraflab; I completely missed Fabian's last comment. :P

prboom2/src/p_plats.c
tpoppins: > Do we want to develop complevel 0 to become more accurate for the sake of demo compatibility with Vanilla Doom 1.2? Or do we want to keep it as it is with known bugs for the sake of demo compatibility with itself?

I overlooked your comment during the holiday season. These are important questions to consider. 

Fortunately, the issue is clear here. The number of -cl0 demos recorded with prboom-plus is insignificant compared to that of Boom- and MBF-format demos - or even to the existing body of v1.2 demos recorded with vanilla - and of those -cl0 demos currently only two or three are known to be broken by improvements in v1.2 emulation. 

So there's no case to make for a "prboom -cl0 standard", this is just a case of a handful broken demos. The will be marked accordingly on DSDA (one or two of mine already are), and perfecting v1.2 emulation should go on unhindered.

These changes may indeed present a significant issue in DSDA-Doom, used to record dozens of Heretic -cl0 demos on DSDA. Should that happen I'm quite confident that kraflab is more than up to the task of coming up with a solution both elegant and efficient.

kraflab: There's no issue for dsda-doom. As I posted in my comment, this logic is already happening in dsda-doom, but only for heretic. I am concerned though if this patch is completely correct, since it _does_ have different cases than heretic's code. It implies heretic patched this switch, but only for some of the plat types? 🤔

tpoppins:
Any outstanding issues with this PR left?

fabiangreffrath:
> Any outstanding issues with this PR left?

Well, it changes complevel behaviour, which we don't take lightly.

fabiangreffrath:
Also, this still needs comparison with the Heretic code, because as @kraflab pointed out, it is improbable that the changes between Doom 1.2 / Heretic and Doom > 1.2 only affected *some* plats types.

tpoppins:
Your caution was justified, Fabian. 

Bob Tausworthe's ABYSS.LMP from [ABYSSL.ZIP](https://archive.org/download/cdrom-1001-nights-doom/1001NightsOfDoom1995wickedSensations.iso/LMPS%2FABYSSL.ZIP) desyncs with this PR. Yet it's fine in my Choco and Crispy forks that adopt the same change.

So what's the problem, then? The problem is that I am a blind dumbass. The relevant sections of p_plats.c  in Choco/Crispy and in prb+ both use a switch statement but the cases are not identical. Bad adaptation of ST's good code on my part. The change indeed should be applied only in _raiseAndChange_ and _raiseToNearestAndChange_ cases. 

```
          // lift types and pure raise types are done at end of up stroke
          // only the perpetual type waits then goes back up
          switch(plat->type)
          {
            case blazeDWUS:
            case downWaitUpStay:
            case genLift:
              P_RemoveActivePlat(plat);     // killough
              break;

            case raiseAndChange:
            case raiseToNearestAndChange:
            // In versions <= v1.2 these platform types always remain active.
            if (compatibility_level > doom_12_compatibility)
            {
              P_RemoveActivePlat(plat);
              break;
            }
            default:
              break;
          }
```
With that corrected the JCM2-16 E2M8 demo and BT's ABYSS.LMP play to the end properly.

The comparison with Heretic could be a red herring. Even though Heretic was based on the v1.2 code there's no guarantee that the code of the respective p_plats.c files would be identical, is there? I suppose the best way to make sure that the above change accurately emulates v1.2 would be to record some demos with the v1.2 vanilla on a few levels that utilize _raiseAndChange_ action specials (## 14, 15, 66 and 67, according to p_switch.c) and see if the resultant demos desync in prb+ with the latest change.

kraflab:
I'm not 100% sure how I feel about these small fixes to the ancient complevels, especially since a demo exists with the wrong behaviour too. At what point is the wrong behaviour actually ok? For instance, if someone digs up demos recorded in actual MBF there would be plenty of desyncs right? I don't know if I would fetch this into dsda-doom 🤔

tpoppins:
Is this PR in limbo?

fabiangreffrath:
> Is this PR in limbo?

I think so. It changes the behavior of an existing complevel, thus turning that complevel incompatible with itself.

tpoppins:
> It changes the behavior of an existing complevel

But so did #33, yet it took less than two days to go through. The issue it fixed was created by me at **your** suggestion, Fabian. 

> thus turning that complevel incompatible with itself.

A neat turn of phrase but it doesn't mean anything. A complevel needs only to be compatible with the engine version it was designed to emulate and with nothing else.

fabiangreffrath:
> But so did #33, yet it took less than two days to go through. The issue it fixed was created by me at **your** suggestion, Fabian.

Erm, sorry, but maybe it wasn't even right of me to suggest this back then.

> A neat turn of phrase but it doesn't mean anything. A complevel needs only to be compatible with the engine version it was designed to emulate and with nothing else.

Sorry, but actually this is not as obviously correct as it should be. During Woof development we (i.e. @rfomin and me) found various occasions in which PrBoom+ complevel 11 behaviour would differ from MBF.exe behaviour in demo-incompatible ways. We have marked these spots in the sources with the `MBF_COMPAT` macro. However, we decided to go the PrBoom+ code paths, in fact turning Woof to be MBF.exe incompatible with its standard build parameters, for the sake of PrBoom+ cl 11 compatibility. 

The reasons are that (1) people use PrBoom+ to record their demos anyway. Nobody uses MBF.exe anymore and even if they did, the engine would crash in at least half of the most famous MBF-compatible megawad's levels because of the nodes formats used there. And (2) people would declare your port broken if it would desync in a complevel 11 demo recorded with PrBoom+ (or desync when played back in that port) because you insist on consistency with the original EXE that nobody cares for anymore.

Things may be a bit different for Vanilla complevels, but even here complevel 2 isn't exactly Doom2.exe v1.9, because it is limit removing and can only emulate Vanilla's quirks and bugs to a certain degree.

So, in the end, we should probably never touch any code again that affects complevel compatibility, at least without a specific `comp[]` guard or similar. I guess the sheer absence of Doom v1.2 demos has prevented us from a bigger impact of the change you pointed out.

tpoppins:
I've been postponing my reply here, gathering evidence and testing; and now that a final release is on the horizon I'd like to have this resolved.

Agreed about MBF compat. Things-not-vanilla fall outside the scope of my interest, all I can say is that by all signs the path you took in Woof! was the  most reasonable approach. Things are indeed quite different with vanilla complevels, though. VERY different. It's a matter of perspective, of course.

I was piqued and somewhat amused by the "sheer absence of ... v1.2 demos" bit, so a couple of months ago I set about corralling every single v1.2 demo I could lay my hands upon. Counting internal DEMOx lumps in PWADs I was hoping to assemble at least 100, with luck maybe up to 200, examples. The final count exceeded my wildest expectations. I have tested and catalogued over 600 v1.2 demos. Of these over 400 reach the exit, making them prime candidates for regression testing; of those almost 200 are bona-fide Maxes (100% kills and secrets). There are about 40 that desync in vanilla; in almost all of the cases it's because the WAD author updated the map after recording the demo (i.e., the demo is older than the WAD).

I'm planning to post the whole collection (including all necessary PWADs) on DW but I haven't tested all of them in latest Choco yet - there are a lot of desyncs. Now the most remarkable part is that with the latest commit from this PR there are no desyncs in prboom-plus (i.e. not more than in vanilla v1.2). Even the desync on the UAC_DEAD "-complevel 0" speedrun recorded with DSDA-Doom I cited earlier turned out to be a false alarm: I accidentally ran it with the first version of the WAD, rather than with the final version found on /idgames.

v1.2 was **the** engine of choice was most of 1994 and some authors continued to use it well into 1995. Countless classics from that era were developed, played and recorded on with it. I think it would not be an exaggeration to state that it is a canonical version. Indeed, there's hardly any comparison with MBF.

v1.2 emulation has been an elusive goal since prboom's early days. Now, 20 years later are we going to stop within the sight of that goal? Let's fix this while we still can.


re: latest commit: I found out that lumping _raiseAndChange_ together with _raiseToNearestAndChange_ as always-active plats was a mistake that breaks a certain switch on ANTHELL4, producing a desync - see [this post](https://github.com/chocolate-doom/chocolate-doom/pull/1207#issuecomment-1317638361) on the "Doom v0.99 demo support" PR in the Choco repo for the details. Incidentally, nukeykt independently arrived at the same conclusion some years before: viz. [the changes](https://github.com/chocolate-doom/chocolate-doom/pull/819/files#diff-f43c851d33fc61b6282c40a4beab2d2d2f1103b71fc2469c1ec7a0f086b9cc0f) to p_plats.c in his "Doom v1.6 emulation" PR.

rfomin:
> I have tested and catalogued over 600 v1.2 demos. Of these over 400 reach the exit, making them prime candidates for regression testing; of those almost 200 are bona-fide Maxes (100% kills and secrets).

This is very impressive, I'm looking forward to testing this collection. Did you find most of these demos on DSDA? How many v1.2 demos do you think were recorded with PrBoom+? We may need to introduce another complevel if there are many.

kraflab:
> ```c
> if (heretic && plat->type == raiseToNearestAndChange) break;
> P_RemoveActivePlat(plat);     // killough
> ```
> 
> 😄 I should have written this down on my list of "why is it different in heretic?" notes when I implemented that! Should it really happen for every case like in this PR or should it mirror this condition from heretic? :^)

So I take it this intuition from the heretic source ended up being right? :^)

kraflab:
Now that this is a tiny change and mirrors the heretic source I'm more inclined to believe it and merge it into dsda-doom. But I'm not actually completely convinced it belongs in pr+. That is, should the final version of pr+ involve a breaking change to demo compatibility, even if it is philosophically sound?

rfomin:
> That is, should the final version of pr+ involve a breaking change to demo compatibility, even if it is philosophically sound?

How about a new complevel for "true" v1.2 emulation?

kraflab:
We wouldn't add a complevel but rather add a new prboom comp error flag if we wanted to go down that path. Those have the function to toggle on wrong behaviour from old pr+ versions. @tpoppins do you want to add one of those? With that then there's probably no reason not to merge this.

kraflab:
I guess it would be something like this:
```c
if (compatibility_level > doom_12_compatibility || prboom_comp[PC_REMOVE_ACTIVE_PLATS])
```

fraansg:
> I've been postponing my reply here, gathering evidence and testing; and now that a final release is on the horizon I'd like to have this resolved.

I understand the possible reasons for wanting to make a final/definitive release of this fork, but I think it should be postponed or planned for a later release (which seems unlikely to me at this point) while PRs like these exist (I don't quit understand why, but it is full of users who refuse to use dsda-doom in favor of this fork).

> I have tested and catalogued over 600 v1.2 demos. Of these over 400 reach the exit, making them prime candidates for regression testing; of those almost 200 are bona-fide Maxes (100% kills and secrets).

It's impressive!

For my part, I think it would be better to wait to see the implications of this PR (and maybe, others?) before the final release.

JadingTsunami:
> I understand the possible reasons for wanting to make a final/definitive release of this fork, but I think it should be postponed or planned for a later release (which seems unlikely to me at this point) while PRs like these exist (I don't quit understand why, but it is full of users who refuse to use dsda-doom in favor of this fork).

I agree, I also don't see the point of a "final release" for this port given it still has active developers maintaining it, responding to bug reports, etc. I also assume it will need maintenance updates now and again.

But perhaps that's a separate thread.

On this one PrBoom comp error flags seem purpose-built for this type of thing; if one is implemented merging this makes sense to me as well.

tpoppins:
>(rfomin) This is very impressive, I'm looking forward to testing this collection. Did you find most of these demos on DSDA?
Most of the demos are culled from various CD-ROMs, all of which are available in [The DOOM Level CD Collection](https://archive.org/details/doom-cds?sort=titleSorter) on archive.org. I don't have the entire collection so there are probably quite a few more demos out there. In particular, I only have a few surviving DHT exam entries and there should be dozens. I hope there might exist a CD-ROM that archived them all before they were nuked off infant sometime in 1994.
Some (a couple dozen) can be still found on /idgames, mostly as internal demos. I didn't use DSDA as the source as there are very few v1.2 demos there (most of them recorded by myself); now that you mention it I can think of one or two I missed (e.g. Mike T's 1_ON_1 max).

If the DSDA crew is interested in preserving some of these runs for historical purposes I could submit about 200 that meet requirements for the Max category. There are also a handful of NM entries and even a couple of UV Speeds.

>(rfomin) How many v1.2 demos do you think were recorded with PrBoom+?
Not more than a dozen. None of them is included in my collection, it is strictly vanilla. In fact, the collection will consist of two parts: the previously mentioned ~600 demos recorded with the real thing on the real hardware (mostly 1994-1995), and about 60 I recorded in the past few months with vanilla under Dosbox and Dosbox-X. These latter have all been extended to display the stats screen at the level's end, all done by hand until I figured out how to hack the vanilla to do it for me. The first part is about 22 MB zipped and includes all the required PWADs. Once I'm done cataloguing and testing the second part it will all be up on DW. And that will include some interesting new hacks of v1.0-1.2 that fix a few demo-related short-sighted design decisions id made early in the Doom development.

Prboom-plus and DSDA-Doom, unlike Choco and Crispy, still write the header in the wrong format with -complevel 0, so the resultant demos are not vanilla-compatible. Note to self about two other outstanding issues: "-fast" and sector special 17 (randomly flickering lights) should be ignored in v1.2 compat. That said, I'm not aware of any prb+/DSDA-D -cl0 demo that would desync after this PR, if you don't count that mongrel -fast v1.2 run on RETURN01 E1M2.

tpoppins:
(kraflab)
> > ```c
> > if (heretic && plat->type == raiseToNearestAndChange) break;
> > P_RemoveActivePlat(plat);     // killough
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > smile I should have written this down on my list of "why is it different in heretic?" notes when I implemented that! Should it really happen for every case like in this PR or should it mirror this condition from heretic? :^)
> 
> So I take it this intuition from the heretic source ended up being right? :^)

Right you are! I must confess I didn't pick up on it at the time, thinking that whatever differences there might be from the Heretic sources could be a red herring since we have no way of knowing exactly what Raven Software did to the v1.2 code - even assuming they got the v1.2 code to work with and not some later, unreleased variation. I still think that, but I admit I underestimated your intuition. :P


> I'm not actually completely convinced it belongs in pr+. That is, should the final version of pr+ involve a breaking change to demo compatibility, even if it is philosophically sound?

Ha, you got your philosophical robe on backwards! ;) What this PR does is fix the last few desyncs on the demos that "-complevel 0" was created to play back in the first place. From this aspect, the change is purely pragmatic. What does this PR break? Show me 600 - heck, even a hundred will do - prb+/DSDA-D -cl0 demos that suddenly go out of sync because of this PR and I will concede that you have a point. At the moment, I fail to see what that point is.  I find the idea of keeping a complevel broken/incomplete just because it has been so for the past 20 years incomprehensible. Now here's where you could apply the "philosophical" epithet, except it would be related to philosophy only as a man in the street perceives it, i.e. something akin to idle theorizing. 

(JadingTsunami)
> I also don't see the point of a "final release" for this port given it still has active developers maintaining it, responding to bug reports, etc. I also assume it will need maintenance updates now and again.

Exactly what I feel, too. Even though I no longer have vested interest in prboom-plus - Crispy has been my go-to port in the past three years - I still would like to see a port with complete v1.2 support.

rfomin:
> Once I'm done cataloguing and testing the second part it will all be up on DW, along with new hacks of v1.0-1.2 that fix a few demo-related short-sighted design decisions id made early in the Doom development.

Did you manage to add statdump support? Maybe it's too complicated, but we could get working [statcheck](https://www.chocolate-doom.org/wiki/index.php/Statcheck).

> Prboom-plus and DSDA-Doom, unlike Choco and Crispy, still write the header in the wrong format with -complevel 0

I tried to fix it here: #564 

So how about using `prboom_comp`?
<details>
  <summary>diff</summary>
  
```diff
diff --git a/prboom2/src/e6y.c b/prboom2/src/e6y.c
index 353b0673..81fc1424 100644
--- a/prboom2/src/e6y.c
+++ b/prboom2/src/e6y.c
@@ -272,6 +272,7 @@ prboom_comp_t prboom_comp[PC_MAX] = {
   {0x00000000, 0x02050007, 0, "-do_not_use_misc12_frame_parameters_in_a_mushroom"},
   {0x00000000, 0x02050102, 0, "-apply_mbf_codepointers_to_any_complevel"},
   {0x00000000, 0x02050104, 0, "-reset_monsterspawner_params_after_loading"},
+  {0x02050104, 0x02060200, 0, "-remove_active_plats"},
 };
 
 void e6y_InitCommandLine(void)
diff --git a/prboom2/src/e6y.h b/prboom2/src/e6y.h
index ed7c86f2..56f9e949 100644
--- a/prboom2/src/e6y.h
+++ b/prboom2/src/e6y.h
@@ -255,6 +255,7 @@ enum
   PC_DO_NOT_USE_MISC12_FRAME_PARAMETERS_IN_A_MUSHROOM,
   PC_APPLY_MBF_CODEPOINTERS_TO_ANY_COMPLEVEL,
   PC_RESET_MONSTERSPAWNER_PARAMS_AFTER_LOADING,
+  PC_v12_REMOVE_ACTIVE_PLATS,
   PC_MAX
 };
 
diff --git a/prboom2/src/p_plats.c b/prboom2/src/p_plats.c
index 2a3c228a..9e612335 100644
--- a/prboom2/src/p_plats.c
+++ b/prboom2/src/p_plats.c
@@ -122,7 +122,7 @@ void T_PlatRaise(plat_t* plat)
 
             case raiseToNearestAndChange:
             // In versions before v1.666 these platform types always remain active.
-            if (compatibility_level > doom_12_compatibility)
+            if (compatibility_level > doom_12_compatibility || prboom_comp[PC_v12_REMOVE_ACTIVE_PLATS].state)
             {
               P_RemoveActivePlat(plat);
               break;

```  
</details>

tpoppins:
👍 for 564.

> So how about using `prboom_comp`?

If I had to go about it this way I'd change
`if (compatibility_level > doom_12_compatibility || prboom_comp[PC_v12_REMOVE_ACTIVE_PLATS].state)`
to
```
if (compatibility_level > doom_12_compatibility || 
    (prboom_comp[PC_v12_REMOVE_ACTIVE_PLATS].state && !demorecording))
```
but what would we do that for? There are no demos with the wrong behavior that this code would accommodate.

I just remembered: there does remain one desync even with this PR: the internal demo on CHURCH. It would need another PR, like [#1529](https://github.com/chocolate-doom/chocolate-doom/pull/1554) on the Choco repo.

rfomin:
> but what would we do that for? There are no demos with the wrong behavior that this code would accommodate.

At some point, the PrBoom+ developers decided to make all changes that affect compatibility optional. Yes, this will not affect any or very few demos, but the same can be said about most of the `prboom_comp` options. Also, I don't think we need to check `demorecording`.

> I just remembered: there does remain one desync even with this PR: the internal demo on CHURCH. It would need another PR, like [#1529](https://github.com/chocolate-doom/chocolate-doom/pull/1554) on the Choco repo.

Of course, I think we should fix this too. Do you want me to do a PR? It's strange that the Choco version hasn't been merged yet.

fabiangreffrath:
I'd agree that this could get merged if there was another config flag that prevents former complevel 0 demos recorded with PrBoom+ to get out of sync.

Also, guys, please don't put too much weight on my idea to publish a "final release". This isn't supposed to mean "you must not use this source port anymore after this". It would just be a symbolic act for me to mark the end of my active maintenance of this port. Because, once this isn't in Debian anymore (and I took care of this 😇 ) chances are very low that I continue to use it regularly.

prboom2/src/e6y.c
```diff
@@ -220,7 +218,7 @@ void e6y_assert(const char *format, ...)
   va_list argptr;
   va_start(argptr,format);
   //if (!f)
-    f = fopen("d:\\a.txt", "ab+");
+    f = M_fopen("d:\\a.txt", "ab+");
```
rfomin, line 221: Irrelevant changes from master ended up in your branch. I guess you staged them by accident?

kraflab:
> but what would we do that for? There are no demos with the wrong behavior that this code would accommodate.

Port versions exist forever. Some people are still running 2.5.1.3. Since people will eternally record demos on old versions, we should support those demos.

prboom2/src/e6y.c
tpoppins: I saw that. The branch is more than a year old. The local file I edited and uploaded is synced to the latest master. I suppose the only way to avoid these false diffs showing up would be to delete my fork and do it all over again.

rfomin: You can revert the last commit, then run `git merge master`, commit the latest changes again, and then push. You will get a branch synced with master. I'm not good with git, maybe there is an easier way.

tpoppins:
>(fabian) chances are very low that I continue to use it regularly.

I've hardly used it myself in the past three years, except for technical stuff like scripts using -levelstat and -skipsec, yet still keep popping up here stirring up trouble every now and then. You did an awesome amount of work here, most importantly making the software renderer usable again, so I hope this is not the last of it.

>(rfomin) Did you manage to add statdump support? Maybe it's too complicated, but we could get working [statcheck](https://www.chocolate-doom.org/wiki/index.php/Statcheck).

No, nothing complex like that, just simple hex hacks using the disassembler output as a guide. _-statcopy_ isn't in v1.2 and earlier, anyway; perhaps implementing it would be possible with kgsws' [ACE Engine](https://www.doomworld.com/forum/topic/130715-ace-engine-dos-doom-ii-vanilla/). Yes, it would be nice to have an automated demo stat tool for the old versions but it is not really necessary. All the necessary data is in the TXT file of the v1.2 demo collection, which was one of the reasons for making it. 

>(kraflab) Since people will eternally record demos on old versions, we should support those demos.

OK, that does make sense.

PR #443, opened 2021-12-24
Do not use sfx_getpow in v1.2 compat

tpoppins:
The first DOOM version that has DSGETPOW (the gong sound played when picking up powerups) present in the IWAD is v1.666.
Credit goes to SmileTheory for the Choco implementation.


fabiangreffrath:
So, it will play the standard "duddut" sound when picking up powerups?

tpoppins:
It will play DSITEMUP.

PR #444, opened 2021-12-25
Fullbright fixes for v1.2 compat

tpoppins:
In versions before v1.4 the Spiderdemon didn't go fullbright while attacking. Note that this means that the whole sprite didn't light up, not to be confused with the (faint) muzzle flashes around its chainguns.

The Soulsphere went fullbright in v1.25 (the Sybex-exclusive shareware). Starting with v1.4 all powerups are rendered fullbright.


tpoppins:
Sorry, will have to resubmit because I forgot to fix the formatting.

PR #451, opened 2022-01-02
Fix bugs with UMAPINFO par times not showing.

### LINKED ISSUE #450: Issues with UMAPINFO partime
I spoke with JadingTsunami on Doomworld, and it appears I've uncovered a couple of UMAPINFO issues related to partime.  First, that attribute is ignored for entries of the style ExMy when x > 4, as I was attempting to create a better UMAPINFO for the main SIGIL.wad file which uses E5My map entries.  Second, I had tried to make one for Plutonia 2, but that was also ignored because of an apparent longstanding issue where only par times found in DeHackEd patches are honored for modified games.

JadingTsunami:
Fix #450 by checking mapinfo when deciding whether or not to display par times.


fabiangreffrath:
Thank you!

PR #456, opened 2022-01-16
Added option to show current game skill on the automap

whouishere:
This adds a new automap widget which shows the current game skill on the automap.
Also added its option to the "automap" setup menu (turned off by default), and and its color (6, red range by default).

I'm not sure about the location of the widget on the automap. Initially I thought about putting it on the right bottom, but I couldn't figure out a way to prevent longer strings to go off screen.


fabiangreffrath:
Is this really something we need to see on the automap all the time once enabled? I have a SKILL cheat in Crispy for this very purpose.

whouishere:
I didn't think of that, it sounds better that way. Definitely will implement it, thanks!

PR #457, opened 2022-01-17
New cheat that shows skill level

whouishere:
This adds a simple cheat which shows the current skill level as a number.

I thought about showing the skill names instead of their corresponding number, but since we don't have any skilltable array like in the equivalent implementation on Crispy Doom, that would mean using a hard-coded table only for this function, which I don't know how acceptable that would be here. Let me know your thoughts about this!

### DIFF
--- a/prboom2/src/m_cheat.c
+++ b/prboom2/src/m_cheat.c
@@ -94,6 +94,7 @@ static void cheat_megaarmour();
 static void cheat_health();
 static void cheat_notarget();
 static void cheat_fly();
+static void cheat_skill();
 static void cheat_comp_ext();
 static void cheat_shorttics();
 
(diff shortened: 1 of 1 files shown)


JadingTsunami:
Rather than keep doing one-off cheats, maybe we should just implement a drop-down console after all. I have one in my personal fork of prboom already and it's useful for things like this.

fabiangreffrath:
> Rather than keep doing one-off cheats, maybe we should just implement a drop-down console after all. I have one in my personal fork of prboom already and it's useful for things like this.

Is it based on this?
https://github.com/kraflab/dsda-doom/blob/master/prboom2/src/dsda/console.c

Let's see what you got, maybe we can merge it!

fabiangreffrath (approved):

JadingTsunami:
> > Rather than keep doing one-off cheats, maybe we should just implement a drop-down console after all. I have one in my personal fork of prboom already and it's useful for things like this.
> 
> Is it based on this? https://github.com/kraflab/dsda-doom/blob/master/prboom2/src/dsda/console.c
> 
> Let's see what you got, maybe we can merge it!

Ha, embarrassingly I did not know about that, but through the magic of [simultaneous invention](https://en.wikipedia.org/wiki/Multiple_discovery) I seem to have made something vaguely similar. I'll post a separate topic on the issue.

PR #466, opened 2022-02-06
Add run as a mouse-settable control (mouseb_speed)

Archenoth:
When setting key bindings in the settings menu, currently you can currently set mouse buttons to quite a few actions--however, run also isn't one of them..!

This PR adds support for setting bindings like this:
The prboom key bindings page with "Run" highlighted, and also with MB2 (Mouse button 2) set as a binding


fabiangreffrath:
Thank you!

PR #467, opened 2022-02-10
clean up timer functions, once again

fabiangreffrath:
This time based on the implementation in Woof!


fabiangreffrath:
@tpoppins May I kindly ask you to post another test build on the DW thread?

tpoppins:
Roger.

kraflab:
The only change here is that we call `SDL_GetTicks()` once instead of twice, right? Why does this change anything?

PR #469, opened 2022-02-10
avoid PATH_MAX redefinition in doomtype.h

### LINKED ISSUE #468: PATH_MAX redefinition in doomtype.h
There is something I do not understand about the redefinition of `PATH_MAX` in the `doomtype.h` file
On WIN32 platform the definition of PATH_MAX *should* be mapped to `MAX_PATH` which is 260.
Also by default `PATH_MAX` is defined to 260 in MinGW headers (might be different on other compilers).

https://github.com/coelckers/prboom-plus/blob/863702b94de593c53d7aaaf871de6d0926ef15ae/prboom2/src/doomtype.h#L68-L86

What happens is that if `window.h` is included before `doomtype.h`, then the macro does not get redefined to 1024 unless you are using MSVC, in this case PATH_MAX is undefined to be sure it gets redefined at 1024.
If the `window.h` is included after `doomtype.h`, then PATH_MAX is always defined first to 1024, then window.h *may* redefine it to 260. I say may, because for sure PATH_MAX did not exist on old MSVC versions (could not test on newer, I just have an old VC++6 CD).

So in the end I get a lot of "redefinition warnings" when compiling with MinGW, and I am always scared that there might be some files using one or the other PATH_MAX definition.

So my questions are:
Why bother redefining PATH_MAX? should it not be always available via the system? This is the case on mingw for instance.
If it is required for some compilers to define PATH_MAX, then redefinition should be based on the system and not always be set to 1024. But how exactly? this is the question.

Of course, in general on Window using larger values for PATH_MAX should not be a trouble, unless definitions get mixed, and some buffer manages to get more than 259 characters.

I hope I am clear, maybe I am missing something.

fabiangreffrath:
Fixes #468, does it @RamonUnch ?

### DIFF
--- a/prboom2/src/doomtype.h
+++ b/prboom2/src/doomtype.h
@@ -61,6 +61,7 @@ typedef unsigned char byte;
 #define BETWEEN(l,u,x) ((l)>(x)?(l):(x)>(u)?(u):(x))
 #endif
 
+#include <limits.h>
 
 /* cph - Wrapper for the long long type, as Win32 used a different name.
  * Except I don't know what to test as it's compiler specific
@@ -120,9 +121,6 @@ typedef unsigned __int64 uint_64_t;
 #else
   #define INLINE inline        /* use standard inline */
 #endif
-
-/* CPhipps - use limits.h instead of depreciated values.h */
-#include <limits.h>
 
 /* cph - move compatibility levels here so we can use them in d_server.c */
 typedef enum {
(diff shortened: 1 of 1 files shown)


RamonUnch:
Thanks for being so fast!
Actually looking closer I think the  `#include <limits.h>` should be moved up (in addition of your modification), because it contains the definition of PATH_MAX on UNIX platforms and also for MinGW.

fabiangreffrath:
That'd be it.

RamonUnch:
Good for me!

PR #470, opened 2022-02-13
Move default data directory to XDG_DATA_HOME on Linux

facespkz:
This takes the current default into account, so it would only be noticeable on a fresh install.

Also a minor refactor for generating the old path.

### DIFF
--- a/prboom2/src/SDL/i_system.c
+++ b/prboom2/src/SDL/i_system.c
@@ -46,7 +46,6 @@
 #include <time.h>
 #include <signal.h>
 #include <string.h>
-#include <dirent.h>
 #ifdef _MSC_VER
 #define    F_OK    0    /* Check for file existence */
 #define    W_OK    2    /* Check for write permission */
@@ -351,7 +350,7 @@ static const char prboom_dir[] = {"prboom-plus"};
 const char *I_DoomExeDir(void)
 {
   static char *base;
-  DIR *data_dir;
+  struct stat data_dir;
 
   if (!base)        // cache multiple requests
     {
@@ -365,10 +364,10 @@ const char *I_DoomExeDir(void)
       base = malloc(p_len);
       snprintf(base, p_len, "%s/.%s", home, prboom_dir);
 
-      data_dir = opendir(base);
-      if (data_dir)
+      stat(base, &data_dir);
+      if (S_ISDIR(data_dir.st_mode))
         {
-          closedir(data_dir);
+          return base;
         }
       else
         {
--- a/prboom2/src/SDL/i_system.c
+++ b/prboom2/src/SDL/i_system.c
@@ -354,14 +354,16 @@ const char *I_DoomExeDir(void)
   if (!base)        // cache multiple requests
     {
       char *home = getenv("HOME");
+      char *p_home = strdup(home);
       size_t len = strlen(home);
       size_t p_len = (len + strlen(prboom_dir) + 3);
 
       // I've had trouble with trailing slashes before...
-      if (home[len-1] == '/') home[len-1] = 0;
+      if (p_home[len-1] == '/') p_home[len-1] = 0;
 
       base = malloc(p_len);
-      snprintf(base, p_len, "%s/.%s", home, prboom_dir);
+      snprintf(base, p_len, "%s/.%s", p_home, prboom_dir);
+      free(p_home);
 
       stat(base, &data_dir);
       if (!S_ISDIR(data_dir.st_mode))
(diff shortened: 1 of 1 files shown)


prboom2/src/SDL/i_system.c
```diff
@@ -46,6 +46,7 @@
 #include <time.h>
 #include <signal.h>
 #include <string.h>
+#include <dirent.h>
```
fabiangreffrath, line 49: There is no dirent.h on Windows. Please use stat() and S_ISDIR instead, c.f. the code in d/main.c:IdentifyVersion().

fabiangreffrath (changes requested):

prboom2/src/SDL/i_system.c
facespkz: Woops. Latest commit should fix that.

prboom2/src/SDL/i_system.c
```diff
@@ -344,22 +344,38 @@ const char* I_GetTempDir(void)
...
       size_t len = strlen(home);
+      size_t p_len = (len + strlen(prboom_dir) + 3);
 
-      base = malloc(len + strlen(prboom_dir) + 1);
-      strcpy(base, home);
       // I've had trouble with trailing slashes before...
-      if (base[len-1] == '/') base[len-1] = 0;
-      strcat(base, prboom_dir);
-      mkdir(base, S_IRUSR | S_IWUSR | S_IXUSR); // Make sure it exists
+      if (home[len-1] == '/') home[len-1] = 0;
```
fabiangreffrath, line 361: "The application shall ensure that it does not modify the string pointed to by the getenv() function."
https://pubs.opengroup.org/onlinepubs/009696899/functions/getenv.html

prboom2/src/SDL/i_system.c
```diff
@@ -344,22 +344,38 @@ const char* I_GetTempDir(void)
...
+          char *prefpath = SDL_GetPrefPath("", prboom_dir);
+          size_t prefsize = strlen(prefpath);
+
+          free(base);
+          base = malloc(prefsize);
+          strcpy(base, prefpath);
+          // SDL_GetPrefPath always returns with trailing slash
+          if (base[prefsize-1] == '/') base[prefsize-1] = 0;
+          SDL_free(prefpath);
+        }
```
fabiangreffrath, line 378: Please keep the `mkdir()` part, the directory returnerd by `SDL_GetPrefPath()` does not necessarily exist already.

prboom2/src/SDL/i_system.c
```diff
@@ -344,22 +344,38 @@ const char* I_GetTempDir(void)
...
+
+      stat(base, &data_dir);
+      if (!S_ISDIR(data_dir.st_mode))
+        {
+          char *prefpath = SDL_GetPrefPath("", prboom_dir);
+          size_t prefsize = strlen(prefpath);
+
+          free(base);
+          base = malloc(prefsize);
+          strcpy(base, prefpath);
```
fabiangreffrath, line 374: How about `strdup()` here?

fabiangreffrath (changes requested):

prboom2/src/SDL/i_system.c
facespkz: SDL Wiki:
> This function will decide the appropriate location in the native filesystem, create the directory if necessary, and return a string of the absolute path to the directory in UTF-8 encoding.

I don't know if calling another `mkdir()` would change anything, but it's better to have redundancy than not.

fabiangreffrath: Alright, I didn't know about the "create the directory if necessary" part.

facespkz: I put it in as a comment, just in case SDL decides to switch things up.

prboom2/src/SDL/i_system.c
facespkz: Yikes, so that's why that was done.

fabiangreffrath (approved):

prboom2/src/SDL/i_system.c
```diff
@@ -344,22 +344,40 @@ const char* I_GetTempDir(void)
...
-      strcat(base, prboom_dir);
-      mkdir(base, S_IRUSR | S_IWUSR | S_IXUSR); // Make sure it exists
+      if (p_home[len-1] == '/') p_home[len-1] = 0;
+
+      base = malloc(p_len);
+      snprintf(base, p_len, "%s/.%s", p_home, prboom_dir);
+      free(p_home);
+
+      stat(base, &data_dir);
+      if (!S_ISDIR(data_dir.st_mode))
```
fabiangreffrath, line 369: Last thing: Could you please add a comment here and document the change you introduced?

facespkz:
For some reason, maybe 5-10% of the time, the code will return the old path, regardless of if it exists or not. Maybe a bug with the filesystem, but I just thought it would wait for the `stat()` to finish.

fabiangreffrath:
Hm, you don't check stat()'s return value. On success, it should return 0. If it doesn't return 0, this should be treated just the same as if the directory doesn't exist. 

```
if (stat(base, &data_dir) || !S_ISDIR(data_dir.st_mode))
```

facespkz:
After running this about 50 times it looks like it works, so I'll follow this up with another PR.

Issue #474, opened 2022-02-14
Problem with partime and endgame


Boryzeo:
If a map defined in the UMAPINFO file uses endcast, endbunny, endpic or endgame = true, then the game will ignore the partime key.
```
MAP MAP03
{
	partime = 60
	endcast = true
}
```
In this example, the par time for MAP03 is defined to be 1 minute, but when the map ends the par time is shown to be 2 minutes, which is the par time for “The Gantlet”.

JadingTsunami:
The responsible line is here:

`src/g_game.c:1646`
```c
	  if (next[0])
	  {
		  G_ValidateMapName(next, &wminfo.nextep, &wminfo.next);
		  wminfo.nextep--;
		  /*...*/
		  wminfo.partime = gamemapinfo->partime;
```

To resolve this, this function would need to be refactored.

Depending on how much refactoring is desired, the function could be mostly rewritten to avoid the `goto` or an else-if clause for mapinfo par time loading could be added around line 1767.

fabiangreffrath:
https://github.com/coelckers/prboom-plus/commit/d97cb0e78558b83e8c33d9a1d0129570f5b7563e

PR #479, opened 2022-02-17
Fix UMAPINFO par times not loading for finale levels.

JadingTsunami:
Proposing a fix for #474

Summary of changes:

* Load par time from UMAPINFO even if no next level is defined
* Skip overwriting par time if it was loaded from UMAPINFO but other parameters are being inherited from the base game.


fabiangreffrath (approved):

fabiangreffrath (approved):
LGTM

fabiangreffrath:
Thank you!

PR #480, opened 2022-02-18
Don't overwrite user-configured custom docdir

vilhelmgray:
We want to provide a default docdir for PrBoom-Plus, but let's not
overwrite the custom docdir a user may provide.


fabiangreffrath:
Could you please add a comment? I know I won't understand this change anymore in a few moths. 😉

vilhelmgray:
I've updated the commit message to matched the discussion we had in https://github.com/coelckers/prboom-plus/commit/daa454465c80231762f66aa2fa2794b547bbaa55

fabiangreffrath (approved):
Top, thanks!

PR #483, opened 2022-03-08
Generate mixer buffer size from sample rate

### LINKED ISSUE #429: (BUG) Chaingun sound slightly out of sync
There's this weird bug that's been in prboom+ since like 2.4 where the rhythm of the chaingun firing just sounds.... off, compared to other source ports. The best way I can describe it is instead of it being like "dudududududududududun", is more like "dudun dududn dudun dudun dudun". I've compared this with Chocolate, Crispy, and Woof, and PrBoom is the odd one out.

facespkz:
Fixes #429, but breaks `slice_samplecount` behavior. Suggested re-usage on line 234.

### DIFF
--- a/prboom2/src/SDL/i_sound.c
+++ b/prboom2/src/SDL/i_sound.c
@@ -227,21 +227,29 @@ static int addsfx(int sfxid, int channel, const unsigned char *data, size_t len)
   return channel;
 }
 
-static int getBufferSize(void)
+static int getSliceSize(void)
 {
-  int limit, i;
+  int limit, n;
 
   if (snd_samplecount) return snd_samplecount;
 
   limit = snd_samplerate / TICRATE;
 
-  // adapted from Chocolate Doom
-  // gets the smallest power of 2, for feeding into SDL
-  for (i = 0;; ++i)
+  // Try all powers of two, not exceeding the limit.
+
+  for (n=0;; ++n)
   {
-    if ((1 << (i + 1)) > limit)
-      return (1 << i);
+    // 2^n <= limit < 2^n+1 ?
+
+    if ((1 << (n + 1)) > limit)
+    {
+      return (1 << n);
+    }
   }
+
+  // Should never happen?
+
+  return 1024;
 }
 
 static void updateSoundParams(int handle, int volume, int seperation, int pitch)
@@ -696,7 +704,7 @@ void I_InitSound(void)
     /* Initialize variables */
     audio_rate = snd_samplerate;
     audio_channels = 2;
-    audio_buffers = getBufferSize();
+    audio_buffers = getSliceSize();
 
     if (Mix_OpenAudioDevice(audio_rate, MIX_DEFAULT_FORMAT, audio_channels, audio_buffers,
                             NULL, SDL_AUDIO_ALLOW_FREQUENCY_CHANGE) < 0)
@@ -726,7 +734,7 @@ void I_InitSound(void)
     audio.format = AUDIO_S16LSB;
 #endif
     audio.channels = 2;
-    audio.samples = getBufferSize();
+    audio.samples = getSliceSize();
     audio.callback = I_UpdateSound;
     if ( SDL_OpenAudio(&audio, NULL) < 0 )
     {
--- a/prboom2/src/SDL/i_sound.c
+++ b/prboom2/src/SDL/i_sound.c
@@ -231,7 +231,8 @@ static int getSliceSize(void)
 {
   int limit, n;
 
-  if (snd_samplecount) return snd_samplecount;
+  if (snd_samplecount >= 32)
+    return snd_samplecount * snd_samplerate / 11025;
 
   limit = snd_samplerate / TICRATE;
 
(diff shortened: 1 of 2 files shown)


prboom2/src/SDL/i_sound.c
```diff
@@ -227,6 +227,23 @@ static int addsfx(int sfxid, int channel, const unsigned char *data, size_t len)
   return channel;
 }
 
+static int getBufferSize(void)
```
fabiangreffrath, line 230: If you are taking the function from Chocolate Doom, why do you rename it and the variables it uses?

prboom2/src/SDL/i_sound.c
```diff
@@ -227,6 +227,23 @@ static int addsfx(int sfxid, int channel, const unsigned char *data, size_t len)
...
+  limit = snd_samplerate / TICRATE;
+
+  // adapted from Chocolate Doom
+  // gets the smallest power of 2, for feeding into SDL
+  for (i = 0;; ++i)
+  {
+    if ((1 << (i + 1)) > limit)
+      return (1 << i);
+  }
+}
```
fabiangreffrath, line 245: Please keep the comments of the original code (and the exit strategy).

prboom2/src/SDL/i_sound.c
```diff
@@ -227,6 +227,23 @@ static int addsfx(int sfxid, int channel, const unsigned char *data, size_t len)
   return channel;
 }
 
+static int getBufferSize(void)
+{
+  int limit, i;
+
+//if (snd_samplecount) return snd_samplecount;
```
fabiangreffrath, line 234: I like this solution. Please change the default value for `snd_samplecount` to `0` in `m_misc.c`.

fabiangreffrath (changes requested):

prboom2/src/SDL/i_sound.c
facespkz: I was going to rewrite it, but then realized that would be silly.

prboom2/src/SDL/i_sound.c
facespkz: Newest commit is (mostly) as is in Chocolate.

prboom2/src/SDL/i_sound.c
```diff
@@ -227,6 +227,31 @@ static int addsfx(int sfxid, int channel, const unsigned char *data, size_t len)
   return channel;
 }
 
+static int getSliceSize(void)
+{
+  int limit, n;
+
+  if (snd_samplecount) return snd_samplecount;
```
fabiangreffrath, line 234: Please do the following:
```
if (snd_samplecount >= 32)
  return snd_samplecount * snd_samplerate / 11025;
```
so users can keep their config if they fine-tuned it with the previous implementation.

fabiangreffrath (changes requested):

prboom2/src/m_misc.c
```diff
@@ -436,7 +436,7 @@ default_t defaults[] =
   {"pitched_sounds",{&pitched_sounds},{0},0,1, // killough 2/21/98
    def_bool,ss_none}, // enables variable pitch in sound effects (from id's original code)
   {"samplerate",{&snd_samplerate},{44100},11025,48000, def_int,ss_none},
-  {"slice_samplecount",{&snd_samplecount},{512},32,8192, def_int,ss_none},
+  {"slice_samplecount",{&snd_samplecount},{0},32,8192, def_int,ss_none},
```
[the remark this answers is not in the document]
facespkz: Is this necessary? The game already resets it back to 0 if it's invalid.

prboom2/src/SDL/i_sound.c
facespkz: Good idea. Since those users have likely moved away from 512, I'll put this condition too:
```
if ((snd_samplecount >= 32) && (snd_samplecount != 512)) 
  return snd_samplecount * snd_samplerate / 11025;
```

prboom2/src/m_misc.c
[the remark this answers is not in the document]
fabiangreffrath: Yes, because we want to explicitly allow to set it `0` and not support this by accident.

prboom2/src/SDL/i_sound.c
fabiangreffrath: No, please don't. We don't need two "di nothing else" cases.

prboom2/src/m_misc.c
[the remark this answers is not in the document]
fabiangreffrath: Also, it's not set back to `0` but merely not changed at all when the config is parsed.

fabiangreffrath (changes requested):

fabiangreffrath (approved):

prboom2/src/m_misc.c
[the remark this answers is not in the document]
facespkz: I guess I could see how that would have repercussions.

prboom2/src/SDL/i_sound.c
facespkz: I admit that it's a hack, but shouldn't we handle the previous default? Or would it be safer to just do this for newer installs?

fabiangreffrath:
Thank you!

PR #486, opened 2022-03-30
Allow uppercase episode shortcuts in UMAPINFO

facespkz:
Fixes in cases like Sigil, which uses an uppercase letter for its episode shortcut.


fabiangreffrath:
@rfomin Do we also need this in Woof?

rfomin:
> @rfomin Do we also need this in Woof?

Yes, we do.

PR #488, opened 2022-04-02
Add complevel hints for Default compatibility level

facespkz:
~~I was thinking about also moving this option to the bottom or something so the new space doesn't look too jarring.~~
Appends numbers to the start of complevels and shortens names where applicable. Addresses #488.


prboom2/src/m_menu.c
```diff
@@ -3300,25 +3300,25 @@ static const char *gen_skillstrings[] = {
...
-  "MBF",
-  "PrBoom 2.03b",
-  "PrBoom 2.1.x",
-  "PrBoom 2.2.x",
-  "PrBoom 2.3.x",
-  "PrBoom 2.4.0",
-  "Latest PrBoom+",
+  "17/Default",
+  " 0/Doom v1.2",
+  "  1/Doom v1.666",
```
fabiangreffrath, line 3305: Uneven spacing?

prboom2/src/m_menu.c
```diff
@@ -3300,25 +3300,25 @@ static const char *gen_skillstrings[] = {
...
+  " 2/Doom/2 v1.9",
+  " 3/Ultimate Doom",
+  " 4/Final Doom",
+  " 5/DosDoom",
+  " 6/TASDoom",
+  " 7/Boom's vanilla",
+  " 8/Boom v2.01",
+  " 9/Boom",
+  "10/LxDoom",
+  " 11/MBF",
```
fabiangreffrath, line 3315: Here as well?

prboom2/src/m_menu.c
facespkz: The 1 character is pretty thin, so double spaces are used to fill in the gaps.

prboom2/src/m_menu.c
facespkz: It's based on the width of the widest number, hence why everything from 12 onwards looks evenly spaced.

fabiangreffrath: The point is, I don't see why these should be aligned at all, because you'll never see more than one at a time.

prboom2/src/m_menu.c
```diff
@@ -3338,7 +3340,7 @@ setup_menu_t gen_settings2[] = { // General Settings screen2
   {"Maximum number of player corpses", S_NUM|S_PRGWARN, m_null, G_X, G_Y+12*8, {"max_player_corpse"}},
   {"Game speed, percentage of normal", S_NUM|S_PRGWARN, m_null, G_X, G_Y+13*8, {"realtic_clock_rate"}},
   {"Default skill level",              S_CHOICE,        m_null, G_X, G_Y+14*8, {"default_skill"}, 0, 0, NULL, gen_skillstrings},
-  {"Default compatibility level",      S_CHOICE,        m_null, G_X, G_Y+15*8, {"default_compatibility_level"}, 0, 0, NULL, &gen_compstrings[1]},
+  {"Default compatibility level",      S_CHOICE,        m_null, G_X-17, G_Y+15*8, {"default_compatibility_level"}, 0, 0, NULL, &gen_compstrings[1]},
```
fabiangreffrath, line 3343: If you leave this line where it was and remove the extra spaces from the strings in the `gen_compstrings[]` array, which entries would exceed the line width?

prboom2/src/m_menu.c
facespkz: It lines up the names with where they previously were, so that it could fit existing names without shortening them.

prboom2/src/m_menu.c
facespkz: Three entries are truncated by doing this: Ultimate Doom, Boom's vanilla, and Latest PrBoom+. These could be shortened, however.

prboom2/src/m_menu.c
fabiangreffrath: Sorry, but I am not going to merge it like this. Prboom+'s menu is already a mess, but having single lines in a paragraph aligned differently is definitely not going to make it any better.

prboom2/src/m_menu.c
fabiangreffrath: Yes, please.

facespkz: Right, then. Could I also change Doom/2 to put the 2 in parentheses?

fabiangreffrath: Sure, why not.

facespkz: Forgot a complevel, turns out that PrBoom 2.03b gets truncated for taking up just a couple more pixels than anything else. Should PrBoom be shortened for all of them?

fabiangreffrath (approved):

prboom2/src/m_menu.c
```diff
@@ -3313,7 +3313,7 @@ static const char *gen_compstrings[] =
   "9/Boom",
   "10/LxDoom",
   "11/MBF",
-  "12/PrBoom 2.03b",
+  "12/PrB 2.03b",
```
fabiangreffrath, line 3316: Wait a minute. This is inconsistent. Could you abbreviate the other "prboom" strings as well?

facespkz: Sure thing. In retrospect, I could've abbreviated every instance as well as PrB+.

fabiangreffrath:
I could live with that.

PR #497, opened 2022-04-18
UMAPINFO: Fix nextsecret inconsistencies

### LINKED ISSUE #496: UMAPINFO: Inconsistent behavior when endgame and nextsecret are present in a map
In the example map, the blue skull switch is a secret exit and the stone switch is a normal exit.

1. Choose the secret exit, and the map will restart
2. Choose the normal exit, and the map will end
3. If you have used the normal exit at least once, start a new game and choose the secret exit, and the game will proceed to map02
4. Remove "intertextsecret" from UMAPINFO and the map will end when either the normal or secret exit is chosen.

Recommend to honor "nextsecret" if present, regardless of "endgame" flags.

[secret.zip](https://github.com/coelckers/prboom-plus/files/8496835/secret.zip)

rfomin:
Fixes #496

### DIFF
--- a/prboom2/src/g_game.c
+++ b/prboom2/src/g_game.c
@@ -1844,7 +1844,7 @@ void G_WorldDone (void)
 
 		  return;
 	  }
-	  else if (gamemapinfo->endpic[0] && (strcmp(gamemapinfo->endpic, "-") != 0))
+	  else if (gamemapinfo->endpic[0] && gamemapinfo->endpic[0] != '-' && !secretexit)
 	  {
 		  // game ends without a status screen.
 		  gameaction = ga_victory;
--- a/prboom2/src/umapinfo.cpp
+++ b/prboom2/src/umapinfo.cpp
@@ -453,7 +453,7 @@ static int ParseStandardProperty(Scanner &scanner, MapEntry *mape)
 		ParseLumpName(scanner, mape->nextsecret);
 		if (!G_ValidateMapName(mape->nextsecret, NULL, NULL))
 		{
-			scanner.ErrorF("Invalid map name %s", mape->nextmap);
+			scanner.ErrorF("Invalid map name %s", mape->nextsecret);
 			return 0;
 		}
 	}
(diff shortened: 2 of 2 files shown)


prboom2/src/g_game.c
```diff
@@ -1844,7 +1844,7 @@ void G_WorldDone (void)
 
 		  return;
 	  }
-	  else if (gamemapinfo->endpic[0] && (strcmp(gamemapinfo->endpic, "-") != 0))
+	  else if (gamemapinfo->endpic[0] && gamemapinfo->endpic[0] != '-' && !secretexit)
```
fabiangreffrath, line 1847: Why did you have to replace `strcmp()` here? It makes a difference if the `gamemapinfo->endpic` string *is* a minus or merely *starts* with a minus.

fabiangreffrath:
Do you think this will require some rewording in the spec?

prboom2/src/g_game.c
rfomin: `'-'` is a special symbol, not user input. This is an implementation detail of PrBoom+'s UMAPINFO.

rfomin:
> Do you think this will require some rewording in the spec?

This particular case is not worth it, I think. It is difficult to make any logic changes in the PrBoom+ repository.

prboom2/src/g_game.c
rfomin: > endpic = "-MYPIC"

In other places we already check only for `[0] == '-'`: https://github.com/coelckers/prboom-plus/blob/784a17d73818cffa4ec3114cef99bcc990d157f4/prboom2/src/e6y.c#L436
https://github.com/coelckers/prboom-plus/blob/784a17d73818cffa4ec3114cef99bcc990d157f4/prboom2/src/g_game.c#L1842

So we need to fix it everywhere?

fabiangreffrath: Ouch! No, of course not. Thanks!

fabiangreffrath (approved):

prboom2/src/g_game.c
rfomin: > Ouch! No, of course not. Thanks!

I thought you'd say we should fix it everywhere :smile: I think we can.

fabiangreffrath: > I thought you'd say we should fix it everywhere 😄 I think we can.

Fine, but not in this PR. 😁 (i.e. not too much at once)

We should decide on one way of checking for `'-'` then, and probably even put it into a macro. I am all for the `strcmp()` approach, because it still leaves lumps with names that merely begin with `'-'` as valid choices.

Issue #207, opened 2021-02-17
[Request] Make -viddump record screen melting effect


MoskovchenkoD:
From what I can assume, viddump records demo tics or their interpolated frames to match specified `cap_fps` while the screen melting effect isn't considered as a part of a demo and thus is not recorded.
However, this effect can improve video quality because it's a familiar Doom feature and it doesn't take long so you would like to skip it.

MoskovchenkoD:
1 year has passed. Probably, it doesn't make sense to keep it open since there are plenty of other issues which have higher priorities.

MoskovchenkoD:
Guess, this commit does what I requested.
https://github.com/vadosnaprimer/prboom-plus/commit/b1ebd6eedf9c31d69164548f69a99b1f894ac6fe

Would anybody please add an option in the settings to turn it on/off?

MoskovchenkoD:
Sorry. I didn't see an active PR is already up:
https://github.com/coelckers/prboom-plus/pull/506

PR #506, opened 2022-06-01
include wipe effect into video capture

### LINKED ISSUE #503: Melt effect is missing in video capture
Tested on a bunch of 2.5.x releases as well as the latest one, default settings, just running

```
prboom-plus -iwad doom.wad -timedemo mydemo.lmp -viddump video.mkv
```

shows instant scene changes where the melt effect is meant to be.

vadosnaprimer:
fixes #503 and #207
draft while people are testing this

# Context

Video dumping code used to exclude the wipe effect, so those trying to have it in the video had to use kkapture or similar tools, which had problems with consistency.

https://github.com/coelckers/prboom-plus/blob/cf609b3b5b80b1fb72150db538ab1a70233107b4/prboom2/src/d_main.c#L493

Adding `I_CaptureFrame();` to the wipe code is probably enough, I just noticed that wipe effect takes 90 frames no matter what, and framerate of the resulting video is twice the `cap_fps`. So setting `cap_fps` to 35 gives what looks like regular DOS video: 70fps with every other frame being a duplicate.

I have no idea how to make the wipe effect respect its original real-time duration and account for the framerate.

https://github.com/coelckers/prboom-plus/blob/cf609b3b5b80b1fb72150db538ab1a70233107b4/prboom2/src/d_main.c#L245

How exactly `done` is determined and what determines `tics` is way above my head.

So if someone who understands the code can help me out here it'd be awesome. If not, this can already be used to get decent video.

ping @andrey-budko I guess?

### DIFF
--- a/prboom2/src/d_main.c
+++ b/prboom2/src/d_main.c
@@ -246,6 +246,10 @@ static void D_Wipe(void)
       I_UpdateNoBlit();
       M_Drawer();                   // menu is drawn even on top of wipes
       I_FinishUpdate();             // page flip or blit buffer
+      if (capturing_video && !doSkip)
+      {
+        I_CaptureFrame();
+      }
     }
   while (!done);
 }
--- a/prboom2/src/d_main.c
+++ b/prboom2/src/d_main.c
@@ -246,7 +246,7 @@ static void D_Wipe(void)
       I_UpdateNoBlit();
       M_Drawer();                   // menu is drawn even on top of wipes
       I_FinishUpdate();             // page flip or blit buffer
-      if (capturing_video && !doSkip)
+      if (capturing_video && !doSkip && cap_wipescreen)
       {
         I_CaptureFrame();
       }
--- a/prboom2/src/i_capture.c
+++ b/prboom2/src/i_capture.c
@@ -73,6 +73,7 @@ const char *cap_tempfile2;
 int cap_remove_tempfiles;
 int cap_fps;
 int cap_frac;
+int cap_wipescreen;
 
 // parses a command with simple printf-style replacements.
 
--- a/prboom2/src/i_capture.h
+++ b/prboom2/src/i_capture.h
@@ -44,6 +44,7 @@ extern const char *cap_tempfile2;
 extern int cap_remove_tempfiles;
 extern int cap_fps;
 extern int cap_frac;
+extern int cap_wipescreen;
 
 // true if we're capturing video
 extern int capturing_video;
--- a/prboom2/src/m_misc.c
+++ b/prboom2/src/m_misc.c
@@ -1067,6 +1067,7 @@ default_t defaults[] =
   {"cap_tempfile2",{NULL, &cap_tempfile2},{0,"temp_v.nut"},UL,UL,def_str,ss_none},
   {"cap_remove_tempfiles", {&cap_remove_tempfiles},{1},0,1,def_bool,ss_none},
   {"cap_fps", {&cap_fps},{60},16,300,def_int,ss_none},
+  {"cap_wipescreen", {&cap_wipescreen},{0},0,1,def_bool,ss_none},
 
   {"Prboom-plus video settings",{NULL},{0},UL,UL,def_none,ss_none},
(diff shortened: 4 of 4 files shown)


fabiangreffrath:
What's the matter with the seemingly unrelated whitespace changes in `f_wipe.c`?

vadosnaprimer:
While figuring out the logic just noticed it's indented with single spaces while the rest of the file uses 2.

vadosnaprimer:
This should be ready now!

fabiangreffrath:
Could you please revert the purely cosmetic changes? Sorry for nit-picking, but this code is surely not meant to be pretty and changes should be kept at a minimum. 😉

kraflab:
I'd recommend adding an option for this (independent of the one for normal gameplay) that defaults to the old behaviour. This is the type of thing that probably doesn't need to be added to the menu but just to the fields in the config file. This change isn't universally desired :^)

fabiangreffrath:
Legitimate objection! Please introduce a config key for this (not necessarily a menu item) and keep it disabled by default.

vadosnaprimer:
> Legitimate objection! Please introduce a config key for this (not necessarily a menu item) and keep it disabled by default.

There's already an option to disable wipe altogether - `render_wipescreen`. Is it really the goal to have wipe on the screen and make it not go to the video?

fabiangreffrath:
> Is it really the goal to have wipe on the screen and make it **not** go to the video?

Yes. In a project like this, people usually don't want any change without the ability to reverse to the old behavior. And that was to have the melt in-game but not in the captured video.

fabiangreffrath:
This should do it:
```patch
diff --git a/prboom2/src/d_main.c b/prboom2/src/d_main.c
index 244a1447..ab3e5d8f 100644
--- a/prboom2/src/d_main.c
+++ b/prboom2/src/d_main.c
@@ -246,7 +246,7 @@ static void D_Wipe(void)
       I_UpdateNoBlit();
       M_Drawer();                   // menu is drawn even on top of wipes
       I_FinishUpdate();             // page flip or blit buffer
-      if (capturing_video && !doSkip)
+      if (capturing_video && !doSkip && cap_wipescreen)
       {
         I_CaptureFrame();
       }
diff --git a/prboom2/src/i_capture.c b/prboom2/src/i_capture.c
index d6b60cf7..70b77bef 100644
--- a/prboom2/src/i_capture.c
+++ b/prboom2/src/i_capture.c
@@ -73,6 +73,7 @@ const char *cap_tempfile2;
 int cap_remove_tempfiles;
 int cap_fps;
 int cap_frac;
+int cap_wipescreen;
 
 // parses a command with simple printf-style replacements.
 
diff --git a/prboom2/src/i_capture.h b/prboom2/src/i_capture.h
index 1197c1c6..9698f4d7 100644
--- a/prboom2/src/i_capture.h
+++ b/prboom2/src/i_capture.h
@@ -44,6 +44,7 @@ extern const char *cap_tempfile2;
 extern int cap_remove_tempfiles;
 extern int cap_fps;
 extern int cap_frac;
+extern int cap_wipescreen;
 
 // true if we're capturing video
 extern int capturing_video;
diff --git a/prboom2/src/m_misc.c b/prboom2/src/m_misc.c
index c2c0c69c..62817b04 100644
--- a/prboom2/src/m_misc.c
+++ b/prboom2/src/m_misc.c
@@ -1067,6 +1067,7 @@ default_t defaults[] =
   {"cap_tempfile2",{NULL, &cap_tempfile2},{0,"temp_v.nut"},UL,UL,def_str,ss_none},
   {"cap_remove_tempfiles", {&cap_remove_tempfiles},{1},0,1,def_bool,ss_none},
   {"cap_fps", {&cap_fps},{60},16,300,def_int,ss_none},
+  {"cap_wipescreen", {&cap_wipescreen},{1},0,1,def_bool,ss_none},
 
   {"Prboom-plus video settings",{NULL},{0},UL,UL,def_none,ss_none},
   {"sdl_video_window_pos", {NULL,&sdl_video_window_pos}, {0,"center"},UL,UL,
```

fabiangreffrath:
Thanks you!

Issue #505, opened 2022-06-01
2s line midtexture vibration


simc2:
If there's a midtexture on a 2s line the texture will shiver or vibrate vertically if behind it there is a upper/lower texture from a moving ceiling/floor plane covering parts of the midtexture. That is when the midtexture is stationary or inmovable relative to the backside ceiling/floor movement.

A 10% speed gif animation of the effect:
![Untitled 4 (1)](https://user-images.githubusercontent.com/27227020/171330003-2601844f-cf78-4292-88ac-075e269fa7d0.gif)

This happens in *Opelgl renderer only* and when the *uncapped framerate is on* (default). Happens in both Prboom-plus, and DSDA-Doom.

Doesn't appear to happen in sw renderer or in opengl when the uncapped framerate is off. Also it doesn't happen to the midtexture on the same side than the moving plane. 

A test wad for map01 Doom2. Should work if PrBoom-plus is run without config, Options > General >Video mode: opengl as the only change.
[texvibra.zip](https://github.com/coelckers/prboom-plus/files/8811167/texvibra.zip)

JadingTsunami:
PR with proposed fix in #507.

fabiangreffrath:
https://github.com/coelckers/prboom-plus/commit/acd8ce1a426e9856026dbb5d4d51da4f0945e3f5

PR #507, opened 2022-06-01
Fix mid-texture judder when upper and middle textures overlap.

JadingTsunami:
Proposed fix for #505.

### DIFF
--- a/prboom2/src/gl_main.c
+++ b/prboom2/src/gl_main.c
@@ -1893,8 +1893,12 @@ void gld_AddWall(seg_t *seg)
       wall.flag = GLDWF_M2S;
       URUL(wall, seg, backseg, linelength);
 
-      wall.vt = (float)((-top + ceiling_height) >> FRACBITS)/(float)wall.gltexture->realtexheight;
-      wall.vb = (float)((-bottom + ceiling_height) >> FRACBITS)/(float)wall.gltexture->realtexheight;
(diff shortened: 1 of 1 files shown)


prboom2/src/gl_main.c
```diff
@@ -1893,8 +1893,12 @@ void gld_AddWall(seg_t *seg)
       wall.flag = GLDWF_M2S;
       URUL(wall, seg, backseg, linelength);
 
-      wall.vt = (float)((-top + ceiling_height) >> FRACBITS)/(float)wall.gltexture->realtexheight;
-      wall.vb = (float)((-bottom + ceiling_height) >> FRACBITS)/(float)wall.gltexture->realtexheight;
+      wall.vt = (float)((-top + ceiling_height))/(float)wall.gltexture->realtexheight;
+      wall.vb = (float)((-bottom + ceiling_height))/(float)wall.gltexture->realtexheight;
+
+      /* Adjust the final float value accounting for the fixed point conversion */
+      wall.vt /= 1<<FRACBITS;
```
fabiangreffrath, line 1900: Isn't this called `FRACUNIT`?

JadingTsunami: Ah, yeah. For some reason it made more sense to me broken out but I think you are right that `FRACUNIT` is better here. Updated.

PR #513, opened 2022-06-21
allow not to quit on demo end

vadosnaprimer:
Incredibly useful if you don't want your video to end right when the demo ends.

For doom demos submitted to tasvideos we have to manually extend the demo by some unknown in advance duration in a hex editor so the entire ending could be encoded. Having a switch (ON by default, so old behavior remains) allows the encoder to end the video when they feel like it without pressure.


kraflab:

Issue #515, opened 2022-07-09
PrBoom doesn't throw error when incorrect command line parameters are supplied.


kvsari:
PrBoom doesn't exit with an error when an incorrect command line parameter is supplied.

This breaks expected command line UI norms. When an incorrect parameter is supplied the program should exit with an error. Instead PrBoom merrily carries on. This is most problematic with the `-record` option.

There have been many times when I wanted to record a playthrough I may mistype the `-record` option or accidentally drop th `-`. PrBoom (running in fullscreen mode) plays as normal and after the play session is done I think a demo has been recorded until I find out it hasn't due to the mistyped command line parameter.

PrBoom should exit with an error in this case.

I have also detected this problem with mistyped `-skill` command line parameter too. PrBoom continues execution when it should exit with an error.

JadingTsunami:
This one of those things that sounds trivial but when you look at the code...oof. The command line argument processing appears to have evolved over many years. There are tons of unrelated entry points for command line arguments that appear to have been added in an ad-hoc, as-needed manner. There are also "clean-up" routines that attempt to make sense of arguments that are supplied out-of-order by adding some implied context. I believe this is primarily to handle "drag and drop" use cases.

I've got one potential start of a solution in a branch [here](https://github.com/JadingTsunami/prboom-plus/tree/error_if_unused_cmdline_args), but it has downsides. Most notably, for an unused argument to be detected, it must start with '-'. Otherwise, we'll need to either untangle the various ad-hoc command line parsers or add a bunch of "set a flag" calls inside them.

Of course we can also have options to ignore unused commands so users can restore the existing functionality if desired (possibly we could have this on by default). I'm sure some shell script or batch file out there will start erroring out if we deployed this as-is.

kraflab:
I've been planning on refactoring command line parameter management in dsda-doom at some point, but this feels like too big of a change to add to pr+. One issue I noticed in that branch is that it doesn't work with parameter overrides - and there's no consistent way to know if they make sense. The current behaviour may not be ideal for all users but it's the way it has been forever, so I don't feel like it's a mistake that necessarily needs fixing right now - given that we are in maintenance mode.

JadingTsunami:
> One issue I noticed in that branch is that it doesn't work with parameter overrides - and there's no consistent way to know if they make sense.

Right, I'm basically side-stepping the issue by looking for '-' at the start of a parameter. Unless you mean something else, in which case, please do elaborate, I'd like to know more.

[...]

PR #517, opened 2022-07-11
Show a message when demo recording starts.

JadingTsunami:
Address #515 by printing a message to the user's view when recording begins.

### DIFF
--- a/prboom2/src/d_deh.c
+++ b/prboom2/src/d_deh.c
@@ -437,6 +437,7 @@ const char *s_STSTR_CHOPPERS     = STSTR_CHOPPERS;
 const char *s_STSTR_CLEV         = STSTR_CLEV;
 const char *s_STSTR_COMPON       = STSTR_COMPON;
 const char *s_STSTR_COMPOFF      = STSTR_COMPOFF;
+const char *s_STSTR_DEMOREC      = STSTR_DEMOREC;
 const char *s_E1TEXT     = E1TEXT;
 const char *s_E2TEXT     = E2TEXT;
 const char *s_E3TEXT     = E3TEXT;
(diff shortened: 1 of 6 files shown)


fabiangreffrath:
As already stated earlier, I like the idea to provide confirmation to the player that a demo is recording. @rfomin What do you say for Woof?

fabiangreffrath (approved):

rfomin:
> As already stated earlier, I like the idea to provide confirmation to the player that a demo is recording. @rfomin What do you say for Woof?

But we have a PR for parameter validation in Woof.

fabiangreffrath:
> But we have a PR for parameter validation in Woof.

Just as an additional confirmation for the user that the current run is recorded. I see this independent from command line validation.

rfomin:
> Just as an additional confirmation for the user that the current run is recorded. I see this independent from command line validation.

Okay, sure.

fabiangreffrath:
@JadingTsunami Could you consider this solution? 

https://github.com/fabiangreffrath/woof/commit/bf2710559c43af7d32ff28516cee2c702d5751fd

JadingTsunami:
> @JadingTsunami Could you consider this solution?
> 
> [fabiangreffrath/woof@bf27105](https://github.com/fabiangreffrath/woof/commit/bf2710559c43af7d32ff28516cee2c702d5751fd)

Sure, I am fine with that as well. The downside of hard-coding the message is that it is not possible to localize it through a patch anymore. But I am not sure anyone would have a strong feeling about it; it may not matter in the end.

JadingTsunami:
Some updates after adopting the proposed approach from @fabiangreffrath.

- `record` will print the demo filename being recorded to.
- `-recordfromto` now shows when recording continues (which is a welcome addition, I think).

PR #519, opened 2022-07-14
Remove shift when initializing translation tables vs. colormap.

JadingTsunami:
As noted in [this post](https://www.doomworld.com/forum/post/2513485), PrBoom+ initializes its color translation tables against a darkened portion of the colormap rather than against the raw base palette. This causes a mismatch between PrBoom+ and other ports' behavior and, arguably, against the reference implementation.

This PR suggests removing the shift to a darker region of the colormap and would bring the translation in line with established norms.

Comment is invited; in particular if anyone has insight into why a darkened portion of the colormap was used.

### DIFF
--- a/prboom2/src/r_draw.c
+++ b/prboom2/src/r_draw.c
@@ -728,9 +728,9 @@ void R_InitTranslationTables (void)
     if (i >= 0x70 && i<= 0x7f)
       {
   // CPhipps - configurable player colours
-        translationtables[i] = colormaps[0][((i&0xf)<<9) + transtocolour[0]];
-        translationtables[i+256] = colormaps[0][((i&0xf)<<9) + transtocolour[1]];
-        translationtables[i+512] = colormaps[0][((i&0xf)<<9) + transtocolour[2]];
+        translationtables[i] = colormaps[0][(i&0xf) + transtocolour[0]];
+        translationtables[i+256] = colormaps[0][(i&0xf) + transtocolour[1]];
+        translationtables[i+512] = colormaps[0][(i&0xf) + transtocolour[2]];
       }
     else  // Keep all other colors as is.
       translationtables[i]=translationtables[i+256]=translationtables[i+512]=i;


JadingTsunami:
> Do we know _why_ it does it differently?

That's the question I have as well. It feels too intentional to be an oversight but too irrelevant to not be a bug at the same time.

In Boom 2.02, it is done like this:

```c
      {   // map green ramp to gray, brown, red
        translationtables[i] = 0x60 + (i&0xf);
        translationtables [i+256] = 0x40 + (i&0xf);
        translationtables [i+512] = 0x20 + (i&0xf);
      }
```

PrBoom 2.02 is the same. By the earliest PrBoom-Plus version I can find, it is as it is above with the shift.

One _might_ infer this was done because the colorized versions of the sprites are not based on hardcoded offsets anymore but instead on the `mapcolor_player` config setting, so the idea was "now that it's configurable, make it halfway in the dark range rather than raw palette remapping." But that may just be telling stories.

I think at this point and in the absence of justified objections, we just treat it as a bug and fix it.

fabiangreffrath (approved):
I agree with @JadingTsunami that Pr+ is really the odd one out here if compared to *any* other source port. Looks like a rather subjective undocumented change to me.

kraflab:
Unfortunately this has fixed red in exchange for breaking some kinds of gray, see here: https://github.com/kraflab/dsda-doom/issues/130

JadingTsunami:
> Unfortunately this has fixed red in exchange for breaking some kinds of gray, see here: [kraflab/dsda-doom#130](https://github.com/kraflab/dsda-doom/issues/130)

Ah, I see. One simple option is, we can change the player arrow color so the offsets work as in the original translation map. Of course we can use a different method altogether that does not rely on the colormap; either seems reasonable.

For my take I'd just revert away from using the colormap for translations as PrBoom+ is an outlier as far as I see and it would only cause more issues.

PR #520, opened 2022-07-27
Formatted code

7dog123:
No code improvemets just Formating C and header filed to look a little more cleaner by using.

astyle --style=allman  --break-blocks --unpad-paren --pad-oper --mode=c ICONS/*.h data/*.c,*.h src/*.c,*.h src/MAC/*.h,*.m src/MUSIC/*.c,*.h src/PCSOUND/*.c,*.h src/POSIX/*.c src/SDL/*.c,*.h src/TEXTSCREEN/*.c,*.h src/WIN/*.c,*.h


kraflab:
These kinds of code-base-wide edits are something we avoid. I understand the desire to clean things up, but there's not much actual reason to in a code base that is in maintenance mode, while the potential for causing chaos in forks is high. The chosen style is also arbitrary.

coelckers:
Indeed. Commits like this make it very hard to update a forked code base beyond it.
And if it has to be done, do it in smaller chunks that are easier to digest.

Issue #101, opened 2020-06-28
Portmidi interpereting pitch bends oddly after listening to certain midis


Alper002:
Steps to recreate: (using [25 years on Earth](https://www.doomworld.com/forum/topic/108938-doom-ii-25-years-on-earth-32-map-limit-removing-old-school-megawad-released/) as an example wad)
1.Use portmidi as midi player
2.Load up the wad normally and use idclev to get to map29
3.Start the game again, this time with -warp 29
4.Compare how the pitch bends in the music sounds

Alternatively, in [Three is a crowd](https://www.doomworld.com/forum/topic/115177-three-is-a-crowd-vanilla-megawad-with-three-protagonists-finally-released/):
1.Use portmidi as midi player
2.Start the game with the wad loaded
3.use idmus10/idclev10
4.use idmus03/idclev03
5.use idmus10/idclev10
6.Compare how MAP10's pitch bends sound before and after playing MAP03's midi

I am assuming the reason there's a discrepancy between starting normally and idcleving compared to using -warp in 25yoe is that the title screen midi has the property that 3iac's MAP03 midi has.

Alper002:
After some searching, I think I have a good guess for why this issue occurs. Pitch bend range changes caused by Registered Parameter Numbers (Which [this page](https://www.recordingblogs.com/wiki/midi-registered-parameter-number-rpn) seems to describe pretty well) aren't reset when a new midi is played. This causes all midis played afterwards to use the same pitch bend range until a new one is used by a midi.

fabiangreffrath:
```
-- a/prboom2/src/MUSIC/portmidiplayer.c
+++ b/prboom2/src/MUSIC/portmidiplayer.c
@@ -387,6 +387,7 @@ static void pm_stop (void)
   int i;
   unsigned long when = Pt_Time ();
   pm_playing = 0;
+  const short norm_pitch = 0x2000;
   
 
   // songs can be stopped at any time, so reset everything
@@ -394,6 +395,7 @@ static void pm_stop (void)
   {
     writeevent (when, MIDI_EVENT_CONTROLLER, i, 123, 0); // all notes off
     writeevent (when, MIDI_EVENT_CONTROLLER, i, 121, 0); // reset all parameters
+    writeevent (when, MIDI_EVENT_PITCH_BEND, i, norm_pitch & 0x7f, (norm_pitch >> 7) & 0x7f); // reset pitch bending
   }
   // abort any partial sysex
   sysexbufflen = 0;
[...]

PR #523, opened 2022-08-19
Fix PortMidi level transitions

ceski-1:
**Bug:**
When a music track changes to a new one, old midi settings carry over if the new track doesn't redefine them. This causes problems with tracks that assume default settings are being used. This was partially addressed in [#101](https://github.com/coelckers/prboom-plus/issues/101) but for pitch bend range only.

**How to recreate:**
Download [Sunlust](https://www.doomworld.com/forum/topic/68089-sunlust-on-idgames/). Launch the game with `prboom-plus.exe -file sunlust.wad -warp 1` and listen to the music. Now launch the game with `prboom-plus.exe -file sunlust.wad`, start a new game manually, and listen again. The celesta (ch. 1 and 2) echoes badly and the slap bass (ch. 5) is positioned far to the right. Both have too much reverb. This is due to the pan, delay (detune), and reverb values carrying over from the title screen music.

**Fix:**
Set the above controllers, along with chorus and other effects, to their default values during a music change. The default values are from the [official midi specifications](https://www.midi.org/specifications).

### DIFF
--- a/prboom2/src/MUSIC/portmidiplayer.c
+++ b/prboom2/src/MUSIC/portmidiplayer.c
@@ -219,7 +219,7 @@ static void pm_shutdown (void)
 
 
 
-static PmEvent event_buffer[6 * 16];
+static PmEvent event_buffer[14 * 16];
 
 static const void *pm_registersong (const void *data, unsigned len)
 {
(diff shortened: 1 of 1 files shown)


fabiangreffrath:
@rfomin could you have a look, please?

rfomin:
> @rfomin could you have a look, please?

LGTM.

fabiangreffrath:
Thanks!

PR #526, opened 2022-09-02
Add proper SysEx support to PortMidi

ceski-1:
Implementing this also has the benefit of fixing a number of related issues.

Changes:
* Process SysEx events correctly from midi files so music sounds as intended
* Send two SysEx messages, one reset and one master volume, and a smaller set of control changes for compatibility with MS GS Wavetable Synth
* Stop function no longer called twice on shutdown
* Volume no longer initializes twice and incorrectly the first time
* Reset midi device on launch, shutdown, and whenever a song changes
* Configure midi device to GS, GM, GM2, or XG modes
* Configure delay after reset for hardware midi devices (e.g. Roland SC-55)
* Fix regression: wads with mid-level music changes no longer cause the game to pause briefly ([test wad](https://github.com/coelckers/prboom-plus/commit/d1fdf95d3476834079400d54049c4ec11ddfbf20))
* Music changes no longer result in occasional "chirp" sound
* Volume levels no longer set two different ways and too low when using the Windows setting "mus_extend_volume"

New settings in prboom-plus.cfg:

`mus_portmidi_reset_type "gs"` (gs, gm, gm2, xg)
Reset type. Default is fine for most users.

`mus_portmidi_reset_delay 0` (0 to 2000 ms)
Delay after reset. Default is fine for software devices like MS GS Wavetable Synth, VirtualMIDISynth, or Roland Sound Canvas VA. For hardware devices like the Roland SC-55, try around 100 to 300 ms.

`mus_portmidi_filter_sysex 0` (0 or 1)
Filter (block) SysEx messages from midi files. Default (0 = don't filter) allows midi files to sound as intended. Pr+ couldn't process SysEx messages before and enabling this setting replicates that behavior at the expense of music accuracy.

### DIFF
--- a/prboom2/src/MUSIC/portmidiplayer.c
+++ b/prboom2/src/MUSIC/portmidiplayer.c
@@ -137,7 +137,7 @@ static void reset_device (unsigned long when)
   else // default to "gs"
     Pm_WriteSysEx(pm_stream, when, gs_reset);
 
-  // additional resets for compatibility
+  // additional resets for compatibility with MS GS Wavetable Synth
   Pm_Write(pm_stream, event_buffer, 13 * 16);
 
   use_reset_delay = mus_portmidi_reset_delay > 0;
--- a/prboom2/src/MUSIC/portmidiplayer.c
+++ b/prboom2/src/MUSIC/portmidiplayer.c
@@ -124,6 +124,7 @@ static unsigned char gs_reset[] = {0xf0, 0x41, 0x10, 0x42, 0x12, 0x40, 0x00, 0x7
 static unsigned char gm_system_on[] = {0xf0, 0x7e, 0x7f, 0x09, 0x01, 0xf7};
 static unsigned char gm2_system_on[] = {0xf0, 0x7e, 0x7f, 0x09, 0x03, 0xf7};
 static unsigned char xg_system_on[] = {0xf0, 0x43, 0x10, 0x4c, 0x00, 0x00, 0x7e, 0x00, 0xf7};
+static PmEvent event_buffer[13 * 16];
 
 static void reset_device (unsigned long when)
 {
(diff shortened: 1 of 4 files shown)


ceski-1:
Here's a capture from Alien Vendetta map04 to help visualize things:
av_map04

From top to bottom:
1. GS Reset from Pr+ PortMidi.
2. Master Volume from Pr+ PortMidi, scaled to music slider.
3. GS Reset from midi file. Redundant but many midi files don't include a reset.
4. Pr+ PortMidi detects GS Reset and reapplies Master Volume.
5. Reverb Level set to 80 from midi file.
6. Reverb Time set to 80 from midi file.
7. Ch. 12 Pan set to 0 from midi file.
8. Voice Reserve tweaks from midi file to avoid notes cutting off.
9. Various channels configured just before notes begin to play.

rfomin:
The problem with this PR is that MS GS Wavetable Synth does not support SysEx messages if I remember correctly. Try to test the examples from here: https://github.com/coelckers/prboom-plus/issues/101#issue-646884534

ceski-1:
MS GS Wavetable Synth supports SysEx but misses some settings when executing a reset. I made some changes to address this and other odd cases that may occur. It passes the pitch bend range tests that you linked and sounds good now. Thank you.

rfomin:
Thank you! This PR is very interesting, I guess I did something wrong when testing SysEx messages when writing `i_winmusic.c` for Woof/Choco/Crispy.

ceski-1:
No problem! I was going to look at those projects next. It's a little trickier without PortMidi though.

rfomin:
> It's a little trickier without PortMidi though.

You need to use [midiOutLongMsg()](https://docs.microsoft.com/en-us/windows/win32/api/mmeapi/nf-mmeapi-midioutlongmsg) and prepare header for it first. It's a bit inconvenient, but shouldn't be difficult. I'll try this weekend too.

prboom2/src/MUSIC/portmidiplayer.c
```diff
@@ -119,6 +119,63 @@ static const char *pm_name (void)
...
+  if (!strcasecmp(mus_portmidi_reset_type, "gm"))
+    Pm_WriteSysEx(pm_stream, when, gm_system_on);
+  else if (!strcasecmp(mus_portmidi_reset_type, "gm2"))
+    Pm_WriteSysEx(pm_stream, when, gm2_system_on);
+  else if (!strcasecmp(mus_portmidi_reset_type, "xg"))
+    Pm_WriteSysEx(pm_stream, when, xg_system_on);
+  else // default to "gs"
+    Pm_WriteSysEx(pm_stream, when, gs_reset);
+
+  // additional resets for compatibility
```
kraflab, line 140: Could you expand this comment to "for compatibility _with xyz_"?

prboom2/src/MUSIC/portmidiplayer.c
```diff
@@ -185,7 +248,7 @@ static void pm_shutdown (void)
   if (pm_stream)
   {
     // stop all sound, in case of hanging notes
-    pm_stop();
+    // pm_stop() is called on shutdown
```
kraflab, line 251: It feels correct to me to call `pm_stop` inside of the shutdown method, even if in practice it can also happen somewhere else. I mean, nothing is stopping us from calling `pm_shutdown` but not shutting down the port, right? Do we need a guard clause in `pm_stop` like `return if not started`? Is it a matter of cleanliness that you are removing the duplication or does it represent a bug?

ceski-1: You're right, there's a risk with someone calling `pm_shutdown` directly and causing hanging notes. Calling `pm_stop` twice means sending another couple hundred control changes plus an extra sysex reset. I don't think this matters even with a hardware midi device, but it's not clean like you've said. How about checking `pm_playing` in `pm_shutdown` ? If it's still `1`, then run `pm_stop` first.

kraflab: Thanks 👍

prboom2/src/MUSIC/portmidiplayer.c
kraflab: Thanks 👍

kraflab:
It looks good to me. Any further concerns @rfomin @fabiangreffrath ?

fabiangreffrath (approved):
LGTM, thanks!

PR #527, opened 2022-09-05
Fix color translation tables to not rely on player map arrow colors.

JadingTsunami:
A suggested fix for https://github.com/kraflab/dsda-doom/issues/130 which does not use the player's map arrow colors as a basis for color translations (which seems like an odd choice to me).

This would bring the behavior of PrBoom+ for color translation in line with other ports.

Comment is invited.

### DIFF
--- a/prboom2/src/am_map.c
+++ b/prboom2/src/am_map.c
@@ -85,7 +85,7 @@ int mapcolor_frnd;    // friendly sprite color
 int mapcolor_enemy;   // enemy sprite color
 int mapcolor_hair;    // crosshair color
 int mapcolor_sngl;    // single player arrow color
-int mapcolor_plyr[4] = { 112, 88, 64, 32 }; // colors for player arrows in multiplayer
+int mapcolor_plyr[4] = { 112, 96, 64, 176 }; // colors for player arrows in multiplayer
 
 //jff 3/9/98 add option to not show secret sectors until entered
 int map_secret_after;
--- a/prboom2/src/r_draw.c
+++ b/prboom2/src/r_draw.c
@@ -704,6 +704,7 @@ void R_InitTranslationTables (void)
   int i, j;
 #define MAXTRANS 3
   byte transtocolour[MAXTRANS];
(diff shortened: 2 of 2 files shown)


kraflab:
Makes sense to me

fabiangreffrath:
Me too! What a weird choice to bind marine suite color to their automap arrow color in the first place. Thanks for figuring this out!

fabiangreffrath:
Could you also change the automap player color as suggested in the original bug report? And I believe there is also a header somewhere which declares `extern byte mapcolor_plyr[]` which is now probably obsolete.

JadingTsunami:
> Could you also change the automap player color as suggested in the original bug report? And I believe there is also a header somewhere which declares `extern byte mapcolor_plyr[]` which is now probably obsolete.

Sure. Updated. This should bring it in line with the original:

https://github.com/id-Software/DOOM/blob/master/linuxdoom-1.10/am_map.c#L1244

```c
#define REDS		(256-5*16)
[...]
#define GREENS		(7*16)
[...]
#define GRAYS		(6*16)
[...]
#define BROWNS		(4*16)
[...]
```

If someone could test/verify this I'd appreciate it as I do not have access to any networking to try it out. Otherwise I'll have to create a test harness later.

kraflab:
You can play back a coop demo to test instead of creating a network game.

JadingTsunami:
> You can play back a coop demo to test instead of creating a network game.

Ah, thanks for the tip. I was able to play back a 4-player demo and the arrows look the same between chocolate and this patch now, plus the remaps look the same as well.

I think perhaps the original idea here was to allow customizable player colors by changing the player arrow color, but it seems it was never implemented. I see no reason not to merge this now and get rid of this presumed-buggy behavior.

fabiangreffrath:
> And I believe there is also a header somewhere which declares `extern byte mapcolor_plyr[]` which is now probably obsolete.

Never mind, it is still used in g_game.c for some reason.

> Sure. Updated. This should bring it in line with the original:

Thank you!

PR #528, opened 2022-09-11
Added option to restart demo on respawn

JokieW:
This adds a new option for demos to be restarted upon respawning, like the "Restart Level/Demo" button does.

That button is super helpful, but I have been doing nightmare runs, which involves a lot of dying. It's easy in the heat of the moment to press "use" instead of "restart" after dying and not realizing you're still recording in the old demo. Or sometimes you try to open a door or something at the same time you die, and you forget to restart for a new demo.

When this setting is off or a demo is being played, respawning will happen the same as before.
When this setting is on and a demo is being recorded, after the player tapped "use" to respawn, it will restart to a new demo, the same as pressing the restart key.


JadingTsunami:
I don't have any personal objection to this functionality, but is there a reason not to use [dsda-doom](https://github.com/kraflab/dsda-doom) for such recordings?

kraflab:
Maybe we should add info to the readme to make it clear that the port is in maintenance mode and not adding new features.

JokieW:
Oh, I didn't know prboom was in maintenance mode. I thought dsda doom was just a fork for speedrunners. I don't really record for competition though, only for myself and my friends, so it didn't cross my mind to check it out.
Sorry about that then. If it can't go in prboom then this pr can be closed. Looks like dsda doesn't have this either so I can try opening it there.

PR #532, opened 2022-09-17
Use non-linear scaling for high volume levels

ceski-1:
As discussed in https://github.com/coelckers/prboom-plus/pull/531, there is a good argument in favor of the music volume slider behaving like vanilla Doom but modified to smoothly transition when approaching maximum volume. This is so the user can perceive differences between higher volume slider settings which was a shortcoming of vanilla Doom. Source ports that prefer staying faithful to vanilla Doom should continue using the vanilla Doom approach. Other than that, both solutions are pretty similar.

Here's a plot I made from the previous PR. PrBoom+ targets the yellow line now:

chart

Here's a capture of the MIDI output as the music volume slider is adjusted from 0 to 15:

```
F0 7F 7F 04 01 00 00 F7    0
F0 7F 7F 04 01 1E 0A F7    1310
F0 7F 7F 04 01 3D 14 F7    2621
F0 7F 7F 04 01 5B 1E F7    3931
F0 7F 7F 04 01 7A 28 F7    5242
F0 7F 7F 04 01 19 33 F7    6553
F0 7F 7F 04 01 37 3D F7    7863
F0 7F 7F 04 01 56 47 F7    9174
F0 7F 7F 04 01 75 51 F7    10485
F0 7F 7F 04 01 13 5C F7    11795
F0 7F 7F 04 01 32 66 F7    13106
F0 7F 7F 04 01 6C 6E F7    14188
F0 7F 7F 04 01 7A 75 F7    15098
F0 7F 7F 04 01 74 7A F7    15732
F0 7F 7F 04 01 03 7E F7    16131
F0 7F 7F 04 01 7F 7F F7    16383
```

So all SysEx messages are correct. As before, if a MIDI file tries to change the master volume, the message will be intercepted and scaled appropriately to the current volume slider and now with a correction factor applied as well.

I'll leave this as a draft so others can provide feedback.

### DIFF
--- a/prboom2/src/MUSIC/portmidiplayer.c
+++ b/prboom2/src/MUSIC/portmidiplayer.c
@@ -120,9 +120,9 @@ static const char *pm_name (void)
 
 // converts master volume from linear scaling to non-linear scaling that
 // closely follows vanilla doom behavior but with a smooth transition when
-// approaching maximum volume. cast value to float and divide by 1000.
-const int volume_correction[] = {1000, 1200, 1200, 1200, 1200, 1200, 1200, 1200,
-                                 1200, 1200, 1200, 1181, 1152, 1108, 1055, 1000};
+// approaching maximum volume
+const float volume_correction[] = {1.000, 1.200, 1.200, 1.200, 1.200, 1.200, 1.200, 1.200,
+                                   1.200, 1.200, 1.200, 1.181, 1.152, 1.108, 1.055, 1.000};
 
 static dboolean use_reset_delay;
 static unsigned char gs_reset[] = {0xf0, 0x41, 0x10, 0x42, 0x12, 0x40, 0x00, 0x7f, 0x00, 0x41, 0xf7};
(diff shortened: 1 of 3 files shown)


prboom2/src/MUSIC/portmidiplayer.c
```diff
@@ -118,6 +118,11 @@ static const char *pm_name (void)
 #include <delayimp.h>
 #endif
 
+// converts master volume from linear scaling to non-linear scaling that
+// closely follows vanilla doom behavior but with a smooth transition when
+// approaching maximum volume. cast value to float and divide by 1000.
+const int volume_correction[] = {1000, 1200, 1200, 1200, 1200, 1200, 1200, 1200,
```
fabiangreffrath, line 124: I still don't get the "cast to float" part. I mean, you are calculating with integers and the result is meant to be an integer, too. If you do all multiplications first and divisions last, the result should be the same, right? 

Apart from this, I am fascinated by your knowledge and expertise of MIDI implementations and would buy about any code change from you. Because, honestly I have no clue about this stuff and what you write just feels right. :grin:

kraflab: Casting to a float here makes the math more decipherable because the related components are connected. Could already have the array be floats with the 1000 divided in though.

fabiangreffrath: > Could already have the array be floats with the 1000 divided in though.

Yep, that's be more straightforward.

ceski-1: Fixed. Bad habit I picked up from a different project.

fabiangreffrath (approved):

kraflab (approved):
Thanks 😎

PR #533, opened 2022-09-23
Adjust PortMidi volume slider to match midiOutSetVolume

ceski-1:
This PR removes the changes of #531 and #532.

Starting from that baseline, music volume is initially scaled linearly:
```
master_volume_out = slider_value / 15 * master_volume_in
```
This is updated to be equivalent to SDL_mixer which uses WinMM `midiOutSetVolume()` which scales the volume non-linearly: 
```
master_volume_out = sqrt(slider_value / 15) * master_volume_in
```
Here's a plot showing gain vs. slider position for a note that's playing at full volume:

volume-changes
(0 dB means no drop in volume, most people won't hear anything by -40 dB, and -60 dB is silent)

Instead of targeting vanilla Doom's [completely broken](https://github.com/chocolate-doom/chocolate-doom/pull/1508#issuecomment-1250443255) music slider, master volume now emulates SDL_mixer but without using `midiOutSetVolume()` which is incompatible with newer versions of Windows.

This update still has the benefits of the previous two PRs. The music slider has a slightly wider usable range, especially in the lower half, and the upper half continues to have perceptible changes when approaching max volume. Users will perceive the volume to be slightly higher across most of the slider range, but the max volume is the same as before.

Thanks to the previous reviewers for asking the right questions that led to this solution.

### DIFF
--- a/prboom2/src/MUSIC/portmidiplayer.c
+++ b/prboom2/src/MUSIC/portmidiplayer.c
@@ -76,6 +76,7 @@ const music_player_t pm_player =
 #include "lprintf.h"
 #include "midifile.h"
 #include "i_sound.h" // for snd_mididev
+#include <math.h>
 
 static midi_event_t **events;
 static int eventpos;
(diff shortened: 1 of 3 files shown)


ceski-1:
I don't intend to make any more volume changes after this PR. The vestigial `mus_extend_volume` setting should be removed at some point, but that is outside the scope of this update.

fabiangreffrath (approved):

PR #537, opened 2022-09-29
UMAPINFO: bring par times fix in line with Woof

fabiangreffrath:


fabiangreffrath:
@rfomin @kraflab plausibility check, please. 😉

kraflab:
I got a bit lost in the discussions about this issue. The only requirement from my side is that all the ports should still sync with demos recorded in ports that predate the fix. Is that the case?

rfomin:
Looks good to me.

> The only requirement from my side is that all the ports should still sync with demos recorded in ports that predate the fix. Is that the case?

Yes, I hope so. But I only tested it with a simple WAD that I made myself.

fabiangreffrath:
This fix does two things:

1. It adds the ability to set par times to the intermission screen that is shown if `endpic` is set but `nointermission` is not. 
2. It tracks if a par time has been set by means of UMAPINFO and makes sure it is shown on the intermission screen.

prboom2/src/g_game.c
```diff
@@ -1658,7 +1671,13 @@ void G_DoCompleted (void)
 		    for (i = 0; i < MAXPLAYERS; i++)
 		      players[i].didsecret = false;
 		  }
+	  }
+
+	  if (next[0] || intermission)
+	  {
 		  wminfo.didsecret = players[consoleplayer].didsecret;
+		  wminfo.partime = gamemapinfo->partime;
```
JadingTsunami, line 1679: There's no guarantee a par time is set, is there? Doesn't this need a check if the par time is nonzero?

fabiangreffrath: > There's no guarantee a par time is set, is there? Doesn't this need a check if the par time is nonzero?

The former code didn't check for this either. We need to have consistent implementations across ports, that's why I reverted nearly all of the previous fix.

fabiangreffrath: NB: It did check in the part that displays the par time, but this is not to question here.

JadingTsunami: Isn't that going to change the behavior in both cases though?

If no par time is set, a 0-time patch would be drawn, wouldn't it? Before it would just be blank.

fabiangreffrath: Good catch! I will change it to
```c
if (wminfo.partime > 0)
  um_pars = true
```

Issue #540, opened 2022-11-09
OpenGL rendering issue


T-Will133:
Found this rendering glitch with OpenGL in DSDA-Doom and also found it broken in PrBoom+ 2.6.2 as well.
![doom00](https://user-images.githubusercontent.com/59149832/200865912-8545fc08-34a5-4891-af56-d39facd9f88a.png)

This is how it looks in Software.
![doom01](https://user-images.githubusercontent.com/59149832/200865916-b0b8b446-0dc3-434f-a7a5-bfbd37daaf4f.png)

T-Will133:
What Remains Map 04 beginning area. 
https://www.doomworld.com/idgames/levels/doom2/v-z/whtrmns

I noticed in the map editor that the sectors with this glitch are set higher than the sky. 

![whtrmns (edit area) at 2022 11 09 11-45-41 977  R4018](https://user-images.githubusercontent.com/59149832/200890358-f6f48809-8c7b-477c-929c-27b9193db2e1.jpg)

T-Will133:
Sounds great.

T-Will133:
Awesome work!

PR #542, opened 2022-11-10
Fixing wall texture bleed when a sector is inside the sky

JadingTsunami:
This is an initial attempt at solving #540. It needs some testing as GL sky rendering is notoriously tricky.

To explain what's going on:

The sky in the GL renderer is more like "the space where stuff _isn't_".
When you move a texture into the sky, the software renderer will clip it at the ceiling intersection. The GL renderer doesn't know about this and will let the texture stretch into the sky.

To fix this, this patch detects the condition and does a few things:
1. Manually clip the wall top at the sky ceiling intersection point (i.e., what the software renderer does)
2. Insert an invisible blocking wall to prevent stuff behind from drawing
3. Adjust the texture offset to compensate for the "chopped" texture

Perhaps you wonder why (2) is not enough: this is because 2 alone will cause z-fighting as both the sky wall and the drawn texture will compete for the same z coordinate. I could nudge the sky wall in the z-buffer but this seems questionable to me and may cause little bleed-overs to happen.

Comment as always is invited. There could be cleaner ways to do this if anyone thinks of one.

### DIFF
--- a/prboom2/src/gl_main.c
+++ b/prboom2/src/gl_main.c
@@ -1614,6 +1614,7 @@ void gld_AddWall(seg_t *seg)
   float lineheight, linelength;
   int rellight = 0;
   int backseg;
+  dboolean fix_sky_bleed = false;
 
   int side = (seg->sidedef == &sides[seg->linedef->sidenum[0]] ? 0 : 1);
   if (linerendered[side][seg->linedef->iLineID] == rendermarker)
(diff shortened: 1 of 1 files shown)


JadingTsunami:
Small update: I ran through my own custom GL test WAD, plus `tests/sky.wad`, plus a few other known-tricky maps. Did not encounter any issues.

fabiangreffrath (approved):

kraflab (approved):
Thanks 😻

PR #543, opened 2022-11-14
Send all notes/sound off prior to SysEx reset

ceski-1:
Fixes https://github.com/kraflab/dsda-doom/issues/153


fabiangreffrath:
Thank you!

PR #544, opened 2022-11-15
Fix MIDI looping when using reset delay

ceski-1:
This only applies to users who enable `mus_portmidi_reset_delay` for hardware MIDI devices.
1. If the song contains a reset message, the reset delay will no longer increase every song loop.
2. If the song doesn't contain a reset message, the reset delay will only occur on first play and not every song loop.


fabiangreffrath (approved):

fabiangreffrath:
Thanks again!

PR #546, opened 2022-11-17
Back to mapcolor_back = 0 by default

fabiangreffrath:


kraflab:
If there are wads that used 247 for the background, they'll now break right?

T-Will133:
If they do break it'll break in Woof and other vanilla-compatible ports. It's probably still safer to revert to mapcolor.

kraflab:
But woof is also 247 default afaik

fabiangreffrath:
> But woof is also 247 default afaik

Yep:
https://github.com/fabiangreffrath/woof/blob/e4ecf2d16322485e666a3e13291b71dcf7ff3c8e/src/m_misc.c#L1580-L1585

fabiangreffrath:
The change dates back to Boom ("black //jff 4/6/98 new black"):

https://github.com/Doom-Utils/historic-ports/blob/3ae2d25740b9d4ea896f0fde0c73f6b112135ce5/m_misc.c#L502-L503

kraflab:
So this has been the default since shortly after the doom source was released... 😅 
I'd rather add basic mapcolor OPTIONS support to pr+ than change this.

kraflab:
@fabiangreffrath let's make a decision about this one. My vote is "do nothing" because that's my default stance with pr+ and this is a limitation of a port that doesn't have OPTIONS lump support. Any boom wad made since '98 would have been made with 247 being the default automap background and I don't think it makes sense to change that now.

fabiangreffrath:
Yep, agree, same here.

Issue #547, opened 2022-11-28
Title music in MAYhem19.wad does not work with PortMidi backend.


rfomin:
@ceski-1 For some reason the GS reset message is muting the MIDI song title in [MAYhem19.wad](https://www.doomworld.com/idgames/levels/doom2/Ports/megawads/mayhem19), when using MS GS Wavetable Synth. Works with `gm_system_on` message.

ceski-1:
For some reason the song uses instrument "variations" (e.g. bank select MSB values > 0) that aren't supported by most Roland devices or MS GS Wavetable Synth. This causes problems I won't get into (search "capital tone fallback" for more details). In GM mode, these bank selects are ignored, so the song plays fine.

Here's are two solutions:
1. Disable sysex reset by default, and make the non-sysex resets a bit more robust to compensate.
2. Emulate capital tone fallback.

Since PrBoom+ is in maintenance mode, I'll go with option 1. I'll make a PR in a bit. Maybe I'll add option 2 to some other projects.

rfomin:
Thanks for explanation. So is the "GM System On" more compatible? Or is it useless for Roland devices?

ceski-1:
"GM System On" is the safest setting but can make songs sound bland if they were written with GS features. GM works fine on Roland devices.

PR #548, opened 2022-11-29
Clean up and fixes for PortMidi

ceski-1:
1. Remove broken and obsolete volume hack: `mus_extend_volume`
2. Disable SysEx reset by default: `mus_portmidi_reset_type ""`
3. Disable SysEx messages from MIDI files by default: `mus_portmidi_filter_sysex 1`
4. Remove resets covered by "reset all controllers"
5. Add additional resets: volume, pan, and bank select MSB/LSB
6. Remove `Pt_Time()` calls during resets
7. Fix const cast warning
8. Calculate volume scale immediately when new volume received
9. Round master volume to nearest integer instead of down
10. Move master volume message to a static variable
11. Add "all sound off" when pausing game to prevent hanging notes
12. Check for escape sequences in SysEx messages
13. Discard partial/incomplete SysEx messages
14. Remove unnecessary `memmove` from SysEx processing
15. Minor whitespace clean up, similar to dsda-doom
16. Clean up `pm_render()` so it's easier to maintain
17. Fixes https://github.com/coelckers/prboom-plus/issues/547

### DIFF
--- a/prboom2/src/MUSIC/portmidiplayer.c
+++ b/prboom2/src/MUSIC/portmidiplayer.c
@@ -94,7 +94,7 @@ static unsigned long trackstart;
 static PortMidiStream *pm_stream;
 
 #define SYSEX_BUFF_SIZE 1024
-static unsigned char sysexbuff[SYSEX_BUFF_SIZE];
+static byte sysexbuff[SYSEX_BUFF_SIZE];
 static int sysexbufflen;
 
 // latency: we're generally writing timestamps slightly in the past (from when the last time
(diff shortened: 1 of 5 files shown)


prboom2/src/MUSIC/portmidiplayer.c
```diff
@@ -463,36 +404,44 @@ static dboolean is_sysex_reset (unsigned char *data)
...
+
   if (len + sysexbufflen > SYSEX_BUFF_SIZE - 1)
   {
-    lprintf (LO_WARN, "portmidiplayer: ignoring large or malformed sysex message\n");
+    // ignore messages that are too long
     sysexbufflen = 0;
     return;
   }
+
+  if (etype == MIDI_EVENT_SYSEX_SPLIT && sysexbufflen == 0)
```
rfomin, line 419: Looks impressive! A few questions:
What happens if we send an escape sequence? Do you have .mid files with `MIDI_EVENT_SYSEX_SPLIT` events?

prboom2/src/MUSIC/portmidiplayer.c
```diff
@@ -463,36 +404,44 @@ static dboolean is_sysex_reset (unsigned char *data)
...
-    sysexbuff[0] = 0xf0; // start of exclusive (SOX) in front
-    sysexbufflen++;
 
-    #ifdef _WIN32
-    if (!mus_extend_volume)
-    #endif
+  // process message if it's complete, otherwise do nothing yet
+  if (sysexbuff[sysexbufflen - 1] == MIDI_EVENT_SYSEX_SPLIT)
+  {
+    if (is_mastervol(sysexbuff, sysexbufflen))
```
rfomin, line 438: Are there files with master volume events? Maybe we should just ignore these events?

prboom2/src/MUSIC/portmidiplayer.c
ceski-1: According to the midi specs, they're used to pass  characters that are not supposed to be sent to the device and is an exploit in the protocol. I guess an example would be sneaking an 0xFF through which would reboot/reset the device but is is simply the meta event status byte in a midi file. Better to drop the hacks, imo.

prboom2/src/MUSIC/portmidiplayer.c
ceski-1: Yes there are. If for some reason a midi file wants to adjust the volume via master volume, it doesn't hurt to scale it accordingly. I know of two more vendor specific master volume messages (mentioned in Roland and Yamaha manuals) but I haven't seen them in midi files. Actually, I could probably add a check for those just to be safe.

ceski-1:
That should be it for this PR.

fabiangreffrath:
> That should be it for this PR.

If you say that... 😉

fabiangreffrath:
Thank you!

PR #549, opened 2022-12-03
Disable boss actions by default in non-vanilla episodes

### LINKED ISSUE #482: SIGIL E5M8 Boss Death flags active
I was recently playing `SIGIL_v1_21.wad` in PrBoom+ 2.6.2 on HNTR, but when I reached the final level, the game unexpectedly cut to the end when I defeated the Spiderdemon boss. I think it was designed for any source port's non-vanilla (>4) episodes to have boss flags inactive by default for their respective final levels, but I believe those flags have been enabled on accident. Using `SIGIL_COMPAT_v1_21.wad` does not have this issue as it has specific DEHACKED flags to disable the Spiderdemon's boss death flags on E3M8.

I have not tested this on HMP or UV, where a Cyberdemon will also spawn, to see how it affects the boss flags.

I have observed this bug in multiple source ports, the main ones being Woof!, Crispy Doom, and PrBoom+. I wasn't sure if it was a shared piece of code between these source ports causing the issue or if they each have unique issues, but I am intending to make unique bug reports for each one.

FozzeY:
Fixes #482.

### DIFF
--- a/prboom2/src/umapinfo.cpp
+++ b/prboom2/src/umapinfo.cpp
@@ -641,11 +641,16 @@ static int ParseMapEntry(Scanner &scanner, MapEntry *val)
 
 	scanner.MustGetIdentifier("map");
 	scanner.MustGetToken(TK_Identifier);
-	if (!G_ValidateMapName(scanner.string, NULL, NULL))
+	int ep;
+	if (!G_ValidateMapName(scanner.string, &ep, NULL))
 	{
 		scanner.ErrorF("Invalid map name %s", scanner.string);
 		return 0;
 	}
+
+	// Disable boss actions by default in non-vanilla episodes.
+	if (ep > 4)
+		val->numbossactions = -1;
 
 	ReplaceString(&val->mapname, scanner.string);
 	scanner.MustGetToken('{');


kraflab:
This is a demo-compatibility-violating change as well as a change to the umapinfo specification. We cannot make such changes lightly.

kraflab:
There are multiple fixes already in the linked issue, so I don't think this needs a code change.

JadingTsunami:
Hmm, I'm not convinced this is an advisable change.

In the case this code addresses, the WAD author has supplied a UMAPINFO but did not supply bossaction(s). It would be a surprise in that case to override game-default behaviors as the UMAPINFO specification is explicit that the default is to defer to the original source code implementation.

I agree with @kraflab; this solution would be a UMAPINFO specification update, plus demo-breaking, and would need to be coordinated across all the ports at the same time. I also don't see the benefit here: if an author supplies a UMAPINFO, why not supply the explicit boss actions (or `clear` them) as intended?

The linked issue is a complaint about the source code behavior, not UMAPINFO, as I see it.

FozzeY:
Fair enough.

PR #551, opened 2022-12-06
Added an INSTALL file for Windows systems.

nounpatroon:


fabiangreffrath:
Thank you for taking the effort to document the build process. BUT: There is already so much outdated information available in the doc/ directory and also in the source root directory - and instead of e.g. updating the INSTALL document, which still refers to autoconf/automake btw, you add a new document specific to using MinGW in MSYS2 on Windows. 

Most parts of the document are pretty generic instructions how to use pacman in MSYS2 or how to call cmake. Some parts are problematic, e.g. recommending to download cmake from the homepage instead of just installing the appropriate package with pacman. Some other parts are plain wrong, e.g. the instructions to pass uild flags to the configure script - which is just not there anymore.

ceski-1:
This could use a small update (64-bit builds, missing packages), but it's still pretty accurate:
https://github.com/coelckers/prboom-plus/wiki/Building-PrBoom-Plus-on-Windows

PR #552, opened 2022-12-12
PortMidi: Use channel volume control, remove extra delay after reset

ceski-1:
1. **Use channel volume control**: Some devices don't support master volume messages. For best compatibility, volume control should be done using channel volume messages.
2. **Remove extra delay after reset**: This is for SysEx resets from the MIDI file only. It's up to the composer to include a delay. If they didn't include one, they didn't intend for the song to play on retro hardware.

### DIFF
--- a/prboom2/src/MUSIC/portmidiplayer.c
+++ b/prboom2/src/MUSIC/portmidiplayer.c
@@ -313,8 +313,6 @@ static const void *pm_registersong (const void *data, unsigned len)
   }
   eventpos = 0;
 
-  // implicit 120BPM (this is correct to spec)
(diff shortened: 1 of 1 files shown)


ceski-1:
That should do it. There's always something I'd like to change but the Woof/Chocolate features are outside the scope of PrBoom+ maintenance.

fabiangreffrath:
Thank you for this! Are there any more changes to be expected to PrBoom+'s MIDI system?

ceski-1:
No, I'm done with PrBoom+. Feel free to ping if any bugs pop up or if there are any questions.

Issue #559, opened 2023-01-25
[Bug] Hanging notes with specific MIDI


NightFright2k19:
With latest dev snapshot, MIDI music playback hangs when reaching the end/loop point with the following MIDI/MUS file taken from the ["NIN Music Doom Wad"](https://www.doomworld.com/files/file/9138-nine-inch-nails-doom-wad/).

[D_DOOM.zip](https://github.com/coelckers/prboom-plus/files/10498672/D_DOOM.zip)

NightFright2k19:
The song appears to play correctly now with latest snapshot build. Thanks a lot!

PR #560, opened 2023-01-25
PortMidi: Fix songs with missing "hold pedal off" events

ceski-1:
Fixes https://github.com/coelckers/prboom-plus/issues/559

### DIFF
--- a/prboom2/src/MUSIC/portmidiplayer.c
+++ b/prboom2/src/MUSIC/portmidiplayer.c
@@ -610,7 +610,10 @@ static void pm_render (void *vdest, unsigned bufflen)
               eventpos = 0;
               // prevent hanging notes (doom2.wad MAP14, MAP22)
               for (int i = 0; i < 16; i++)
+              {
                 writeevent (when, 0xB0, i, 0x7B, 0x00); // all notes off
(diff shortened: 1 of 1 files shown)


fabiangreffrath:
Thanks for the prompt fix!

prboom2/src/MUSIC/portmidiplayer.c
```diff
@@ -610,7 +610,10 @@ static void pm_render (void *vdest, unsigned bufflen)
               eventpos = 0;
               // prevent hanging notes (doom2.wad MAP14, MAP22)
               for (int i = 0; i < 16; i++)
+              {
                 writeevent (when, 0xB0, i, 0x7B, 0x00); // all notes off
+                writeevent (when, 0xB0, i, 0x40, 0x00); // hold1 (damper/sustain) pedal
```
rfomin, line 615: Should we just reset device on song loop? It seems `i_winmusic.c` is also affected.

ceski-1: The MIDI assoc. only provides a "recommended practice" (RP-015) for how "Reset All Controllers" is supposed to behave so I don't want to risk breaking seamless looping if a device chooses to immediately silences all notes, or takes a long time to execute the event, etc.

PR #564, opened 2023-02-07
Fix incorrect demo header for complevel 0

rfomin:
`complevel 0` demos recorded with PrBoom+ now play correctly in vanilla Doom v1.2.
Thanks to @tpoppins for the bug report.


kraflab:
Error flags are for ensuring proper playback. Don't need them for creating the demos in the first place.

fabiangreffrath (approved):

kraflab:
Mentioned on discord but I'll add here for posterity: since we can't guarantee compatibility with 1.2 and it's really just "prboom+'s 1.2 flavour", it might be better to leave the "wrong" header anyway.

rfomin:
> since we can't guarantee compatibility with 1.2 and it's really just "prboom+'s 1.2 flavour"

Why? With @tpoppins collection of demos, we can pretty much guarantee this, at least with complevel 11 quality. I can implement the missed changes, the `prboom_comp` option and test it.

PR #566, opened 2023-02-07
Disable -fast for 1.2 compat

tpoppins:
The first appearance of the "-fast" string in DOOM.EXE is in the v1.4 beta leak (1994-04-08). Vanilla v1.2 ignores the "-fast" switch.

### DIFF
--- a/prboom2/src/d_main.c
+++ b/prboom2/src/d_main.c
@@ -1655,7 +1655,9 @@ static void D_DoomMainSetup(void)
   // jff 1/24/98 set both working and command line value of play parms
   nomonsters = clnomonsters = M_CheckParm ("-nomonsters");
   respawnparm = clrespawnparm = M_CheckParm ("-respawn");
-  fastparm = clfastparm = M_CheckParm ("-fast");
+  // -fast was not an option until v1.4
+  if (compatibility_level > doom_12_compatibility)
+    fastparm = clfastparm = M_CheckParm ("-fast");
   // jff 1/24/98 end of set to both working and command line value
 
   devparm = M_CheckParm ("-devparm");


kraflab:
This is inconsistent with how other ports handle old demo formats. If someone wants to record -fast in a wad that requires cl0 then that's their choice and it will not play back on the original exe. They can also make solo net demos that don't play back in vanilla already

tpoppins:
I'm firmly of opinion that **vanilla**-format demos recorded on **vanilla** WADs should be playable in **vanilla**.  If they are not, they are broken. Players who wish to record broken demos can do so with old port versions. There's a clear difference between accommodating existing broken demos and encouraging making more of them.

Choco and its forks are the only other ports that support old-format demos, and there's the same PR pending on the Choco repo that should take care of this.

> They can also make solo net demos that don't play back in vanilla already

They don't? Try `ipxsetup -nodes 1`, this "trick" is as old as v1.2.

fraansg:
I understand that it is possible to use the -fast parameter in doom v1.2, but together with -devparm and other configurations (I don't know if -wart or something else) and DoomWiki at least confirms that. The question I have is if there is some kind of official release document that indicates how to use the -fast parameter in v1.2?

On the other hand, I agree with what was pointed out by Kraflab. Unlike Chocolate-Doom this project should not aim for vanilla perfect emulation, but rather maintain backward compatibility of demo-playback.

> I'm firmly of opinion that vanilla-format demos recorded on vanilla WADs should be playable in vanilla. 

If that position were to be taken, how do you distinguish between vanilla and limit-removing wads? What would happen to the vast majority of boom wads that are impossible to play on the original boom.exe?

I think that rather than modifying the behavior of the port to emulate the vanilla exe, we should keep the approach used by prboom so far which has always been about functionality.

kraflab:
> I'm firmly of opinion that **vanilla**-format demos recorded on **vanilla** WADs should be playable in **vanilla**. If they are not, they are broken.

It's a nice opinion but it's not relevant here because prboom+ doesn't follow this rule.

PR #567, opened 2023-02-07
Disable sector special 17 in v1.2 compat

tpoppins:
The first vanilla exec to support sector special 17 (random flicker) was the official v1.4 beta. Disabling it for v1.2 compat solves the desync on DEMO1 of CHURCH.WAD.

### DIFF
--- a/prboom2/src/p_spec.c
+++ b/prboom2/src/p_spec.c
@@ -2667,7 +2667,9 @@ void P_SpawnSpecials (void)
 
       case 17:
         // fire flickering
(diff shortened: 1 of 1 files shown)


kraflab:
This one needs a prboom comp flag as well.

kraflab:
@tpoppins tbh if you want to make this stuff simpler, it might be worth combining your fixed into a single PR and just adding one flag standing for pre-2.6.3 v1.2 behaviour

tpoppins:
Thank you for the advice.
In fact, there's a still simpler solution.

PR #569, opened 2023-02-13
Fix string overflow with large map numbers.

### LINKED ISSUE #568: Maps >100 crash
The UMAPINFO spec [gives "MAP100"](https://github.com/kraflab/umapinfo/blob/master/docs/spec.md#map-entry) as an example, but defining maps >99 will crash both PrBoom+ and dsda-doom.

WAD entry limits are 8 characters. Map markers consume either 2 (ExMy format) or 3 (MAPxx format) characters from that total. GL nodes prefix the map marker with `GL_` which drops another 3 characters.

Options:

1. Cap MAPxx at 99 and ExMy at `len(x)+len(y) <= 3`
2. Skip GL nodes in maps that don't fit the criteria from (1) and cap MAPxx at 99999 and ExMy at `len(x)+len(y) <= 6`
3. Use the limits from (2) for UDMF and limits from (1) for "legacy" map types.

Posting here for discussion. Fixing the crash is trivial but has implications for the spec.

JadingTsunami:
Fixes #568

### DIFF
--- a/prboom2/src/p_setup.c
+++ b/prboom2/src/p_setup.c
@@ -2641,13 +2641,13 @@ void P_SetupLevel(int episode, int map, int playermask, skill_t skill)
   // find map name
   if (gamemode == commercial)
   {
-    sprintf(lumpname, "map%02d", map);           // killough 1/24/98: simplify
-    sprintf(gl_lumpname, "gl_map%02d", map);    // figgi
+    snprintf(lumpname, 9, "map%02d", map);           // killough 1/24/98: simplify
+    snprintf(gl_lumpname, 9, "gl_map%02d", map);    // figgi
   }
   else
   {
-    sprintf(lumpname, "E%dM%d", episode, map);   // killough 1/24/98: simplify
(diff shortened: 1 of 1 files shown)


prboom2/src/p_setup.c
```diff
@@ -2641,13 +2641,13 @@ void P_SetupLevel(int episode, int map, int playermask, skill_t skill)
   // find map name
   if (gamemode == commercial)
   {
-    sprintf(lumpname, "map%02d", map);           // killough 1/24/98: simplify
-    sprintf(gl_lumpname, "gl_map%02d", map);    // figgi
+    snprintf(lumpname, 9, "map%02d", map);           // killough 1/24/98: simplify
```
fabiangreffrath, line 2644: Pedantic, but could you use `sizeof()` please?

JadingTsunami: Sure, I'll do it that way. A new version has been added with that change applied.

fabiangreffrath:
Thank you!

